PR #1, opened 2024-06-03
Encalve Spec

auryn-macmillan:
This PR specifies the Enclave protocol in a [suite of unit tests](https://github.com/gnosisguild/enclave/blob/draft-spec/packages/evm/test/encalve/Enclave.spec.ts), along with implementing Mock contracts for each of the various modular components.


packages/evm/test/enclave/Enclave.spec.ts
```diff
@@ -75,17 +66,16 @@ describe("Enclave", function () {
 
   describe("constructor / initialize()", function () {
     it("correctly sets owner", async function () {
-      // uses the fixture deployment with this.owner set as the owner
       const owner1 = await this.enclave.owner();
-      // create a new deployment with this.otherAccount as the owner
-      // note that this.owner is msg.sender in both cases
-      const enclave = await this.Enclave.deploy(this.otherAccount.address, this.otherAccount.address, this.maxDuration);
-      const owner2 = await enclave.owner();
+      const [, , , otherOwner, otherRegistry] = await ethers.getSigners();
```
auryn-macmillan: Should we just add this to the fixture, since it is repeated a few times already?

PR #2, opened 2024-06-07
Registry Spec

auryn-macmillan:
This PR is a first attempt at specifying the interface for the cyphernode registry, along with its components. It incides mainly in the Committee Selection Pahse

Its main focus is the two-step asynchronous committee selection phase that utilizes a pluggable entity called `IRegistryFilter`.

## Key Points
- **Committee Request**: When a committee is requested, the registry signals the `IRegistryFilter` to initiate the selection of the committee.
- **Committee Publishing**: Later, the filter reports back to the registry, publishing the sharedPublickKey and the list of nodes.
- **Time Limit**: There is a specified time limit for the transition to occur. If the transition does not happen within this time frame, the process is declared stale and unfit to proceed to the activation stage.

## Security Considerations
- The registry provides each filter with a function to determine if a node is fit for inclusion. This allows the filter to implement its sortition function with minimal knowledge of protocol internals (e.g., slashing, availability, duties).
- The registry is responsible for validating each node to be included in the committee. Only enabled and fit nodes are allowed. This extra validation is done inside the publish function. This step becomes necessary because we are for handling adding a filter as permissionless action

## Implementation Considerations
- The key generation process is treated as a black box. Currently, it is modeled as a centralized entity responsible for correctly generating and posting the key. Or in other words, some centralized actor will own a the IRegistryFilter contract used for the sortition.

## Cyphernode State
- This concerns staking, slashing, and outstanding duties per node.
- It will be addressed in a subsequent branch, which will introduce the initial interface for the contract managing Cyphernode deposits.


packages/evm/contracts/interfaces/ICyphernodeRegistry.sol
```diff
@@ -4,48 +4,40 @@ pragma solidity >=0.8.26;
...
-    event NodeAdded(uint256 indexed nodeId, address indexed node);
+    /// @notice This event MUST be emitted when a filter is added to the registry.
+    event FilterAdded(address indexed filter);
 
-    /// @notice This event MUST be emitted when a node is removed from the registry.
-    /// @param nodeId ID of the node.
-    /// @param node Address of the node.
-    event NodeRemoved(uint256 indexed nodeId, address indexed node);
+    /// @notice This event MUST be emitted when a filter is removed from the registry.
+    event FilterRemoved(address indexed filter);
```
auryn-macmillan, line 24: I wonder if filter removal is necessary?
Given that filters are passed in by the requestor, it seems unnessary to need to explicityly remove them. Requestors could simply stop requesting them.

packages/evm/contracts/registry/NaiveCommitteeCoordinator.sol
```diff
@@ -0,0 +1,142 @@
...
+        address[] nodes;
+        uint32[2] threshold;
+        bytes publicKey;
+    }
+
+    struct Node {
+        bool eligible;
+        // Number of duties the node has not yet completed.
+        // Incremented each time a duty is added, decremented each time a duty is completed.
+        uint256 outstandingDuties;
```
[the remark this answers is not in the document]
auryn-macmillan: In either case though, I don't think this can live in the filter contracts. It probably needs to be a core feature of the registry.

packages/evm/contracts/registry/NaiveCommitteeCoordinator.sol
auryn-macmillan, line 1: It appears to me that ICommitteeCoordinator is actually acting as the registry, which seems to be conflating the role of the registry and the filers.

The registry is a core component to the protocol. Global node membership is not something that should be defined by the requestor. Rather, requestors should be able to define the filters that they want to apply to the global registry.

packages/evm/contracts/interfaces/ICommitteeCoordinator.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.8.26;
+
+interface ICommitteeCoordinator {
```
auryn-macmillan, line 4: It seems like this should perhaps be called `IRegistryFilter`?

packages/evm/contracts/registry/NaiveCommitteeCoordinator.sol
auryn-macmillan: Put another way, the `ICyphernodeRegistry` should handle node registration and decommissioning (currently permissioned by an owner address, in future permissionless via staking). Regardless of the filters applied, every node selected for a computation must be a member of the global list of eligible nodes to ensure that they are staked and the system maintains economic security.

packages/evm/contracts/interfaces/ICyphernodeRegistry.sol
cristovaoth: Right, agreed. It crossed my mind, and I kept it in because I thought it might be useful for the registry to exclude some filters in an emergency situation. I will remove it.

packages/evm/contracts/interfaces/ICyphernodeRegistry.sol
```diff
@@ -4,48 +4,40 @@ pragma solidity >=0.8.26;
...
-        bytes publicKey
-    );
+    event CommitteeSelected(uint256 indexed e3Id, bytes publicKey);
 
-    /// @notice This event MUST be emitted when a node is added to the registry.
-    /// @param nodeId ID of the node.
-    /// @param node Address of the node.
-    event NodeAdded(uint256 indexed nodeId, address indexed node);
+    /// @notice This event MUST be emitted when a filter is added to the registry.
+    event FilterAdded(address indexed filter);
```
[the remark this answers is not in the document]
cristovaoth: Right. Yeah we should keep it permissionless. Explicitly adding and emitting the event feels right somehow. What point out is scenario.

You have any concrete thoughts on whether include/remove?

packages/evm/contracts/interfaces/ICommitteeCoordinator.sol
cristovaoth: Sounds good. This was in any case, just a placeholder name, I was counting in advance on your input for the rename

packages/evm/contracts/registry/NaiveCommitteeCoordinator.sol
cristovaoth: Yeah, this is a good point, and it's what I meant when I mentioned in the DMs that I hadn't given any thought to the slashing logic and mechanism. Let me think it through

packages/evm/contracts/interfaces/ICyphernodeRegistry.sol
auryn-macmillan: Yeah, I was wondering about this as well.

My worry here is that it implies an untenable governance risk.

packages/evm/contracts/registry/NaiveCommitteeCoordinator.sol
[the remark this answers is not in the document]
cristovaoth: Initially, I took this comment as a major flaw in the direction of this PR. However, upon revisiting it, I think this is only partially the case.

One thing is certain: the commissioning and decommissioning of nodes do not belong in the filter logic. There’s no question about that, you comment is correct.

However, the Registry already maintains a state indicating whether a Cyphernode is active. For the registry, the only important factor is whether a node can be included in a committee selection. It doesn’t matter whether the node hasn’t been added yet, is serving too many duties, or has been slashed and its bond no longer suffices.

So, the relevant question for the registry is whether a node is fit to be included in an upcoming commission. This is a simple boolean value.

What **seems** to be missing is a new entity to handle the logic for staking, entering commissions, leaving commissions with prizes, and leaving commissions with slashes.

I'm going to push a couple more commits and summarize the current status of the phase design in a consolidated PR comment. I'll try to incorporate this idea there.

Let me know if you have any immediate feedback on this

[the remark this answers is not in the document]
auryn-macmillan: So what you're suggesting here is essentially breaking the registry up into a multiple components, one that is the canonical source of truth for whether or not a given node is currently eligible to be selected for committee and one that is responsible for maintaining the list of eligible nodes (along with slashing conditions, etc).

PR #120, opened 2024-10-01
feat: store hashes, emit keys

auryn-macmillan:
This PR modifies the way node committee public keys are handled. Rather than attempting to put public keys in storage, only a hash of the key is stored. The full key is still emitted as event.

This PR also reverts a previous change where only the hash of the plaintext output was stored. Now the full plaintext ourput is stored.

Not that this PR changes the interface for `Enclave.activate()`, which now requires the committee's public key as input. The hash of the given key is compared with the hash stored in the ciphernodeRegistry.

## Release Notes

- **New Features**
	- Enhanced public key management by requiring hashes instead of raw keys for improved security.
	- Updated `activate` function to include public key as a parameter.
	- Introduced more detailed event emissions for published data.
	- Added new parameters for E3 activation and publishing tasks to improve data handling.

- **Bug Fixes**
	- Clarified conditions for checking published outputs to improve reliability.

- **Documentation**
	- Updated function documentation to reflect changes in parameter types and expected values.

- **Tests**
	- Refined test cases to align with new logic, ensuring accuracy in public key handling and activation processes.


hmzakhalid (approved):

ryardley (approved):
🚀

PR #183, opened 2024-11-17
feat: initial Kad node setup

maschad:
## Description

This is the initial node setup to use Kademlia in a gossipsub config.

## What's Changed

- I've removed the rendevous server
- I've added connection limits, updated the gossipsub config and services needed for Kademlia

## Notes

- Ideally you will have some bootstrap nodes that all nodes can always connect to (at least in the inchoate stages of the network)
- This needs to be tested in a deployed devnet


ryardley:
@maschad I can't seem to push to this branch so I might merge it to a new branch in a new PR.

PR #231, opened 2024-12-30
Add Network Peer actor 

hmzakhalid:
## Summary by CodeRabbit

Based on the comprehensive summary, here are the updated release notes:

- **Infrastructure Updates**
  - Updated GitHub Actions workflow cache mechanism from v2 to v4.

- **Deployment Configuration**
  - Updated peer network configuration for nodes and aggregator.
  - Removed redundant peer entries in deployment configurations.

- **Network Management**
  - Introduced new `DialerActor` for improved peer connection management.
  - Enhanced event bus and message routing for network communications.
  - Migrated from `tokio` to `actix` runtime for network operations.

- **Code Refactoring**
  - Simplified import statements and removed duplicate code.
  - Improved variable naming for clarity.
  - Centralized event type handling across components.

These changes primarily focus on improving the network infrastructure, event handling, and overall code organization.


packages/ciphernode/net/src/network_manager.rs
```diff
@@ -154,16 +128,11 @@ impl Handler<EnclaveEvent> for NetworkManager {
...
-                        .send(NetworkPeerCommand::GossipPublish {
-                            topic,
-                            data,
-                            correlation_id: CorrelationId::new(),
-                        })
-                        .await
-                    {
-                        error!(error=?e, "Error sending bytes to libp2p");
-                    };
+                    peer.send(NetworkPeerCommand::GossipPublish {
```
ryardley, line 131: `send` in actix is async and returns a result. Did you mean to use `do_send()`?

packages/ciphernode/net/src/retry.rs
```diff
@@ -1,70 +0,0 @@
...
-/// Retries an async operation with exponential backoff
-///
-/// # Arguments
-/// * `operation` - Async function to retry
-/// * `max_attempts` - Maximum number of retry attempts
-/// * `initial_delay_ms` - Initial delay between retries in milliseconds
-///
-/// # Returns
-/// * `Result<()>` - Ok if the operation succeeded, Err if all retries failed
-pub async fn retry_with_backoff<F, Fut>(
```
ryardley, line 27: I think this is going to be really useful as a function as we go to harden up our actors. Eg. ensuring our set and forget  database writes actually happen so I think it is better to attempt to use this function as opposed to bake it in to the dialer.

packages/ciphernode/net/src/network_peer.rs
```diff
@@ -29,135 +51,254 @@ pub struct NodeBehaviour {
...
+        Ok(())
     }
+}
+
+impl Actor for NetworkPeer {
+    type Context = Context<Self>;
+
+    fn started(&mut self, ctx: &mut Self::Context) {
+        // Setup stream handling for swarm events
+        ctx.add_stream(Box::pin(futures::stream::unfold(
```
ryardley, line 176: cool!

packages/ciphernode/net/src/network_peer.rs
```diff
@@ -29,135 +51,254 @@ pub struct NodeBehaviour {
...
+            &mut self.swarm,
+            |swarm| async move { Some((swarm.select_next_some().await, swarm)) },
+        )));
 
-    pub fn rx(&mut self) -> broadcast::Receiver<NetworkPeerEvent> {
-        self.event_tx.subscribe()
+        if let Some(dialer) = &mut self.dialer {
+            dialer.send(SetNetworkPeer(ctx.address()));
+            if !self.peers.is_empty() {
+                dialer.do_send(DialPeers(self.peers.clone()));
```
ryardley, line 184: It might be simpler to spawned `self.peers.len()` dialers for each peer in the peer list with a dialer being a short lived actor in charge of dialing a specific address and handling expbackoff etc. having the dialer handle multiple requests seems more complex than it needs to be?

packages/ciphernode/net/src/network_peer.rs
hmzakhalid: yeah but sadly not possible, have to resort to polling.

packages/ciphernode/net/src/network_peer.rs
hmzakhalid: The issue with that approach is passing events between the dialer and the peer. In that case, we would have to manage all the dialers inside the NetworkPeer and loop through the dialers for swarm events to send the event to the appropriate dialer. Although, we could create a separate EventBus for NetworkPeerEvent that the dialers can subscribe to in order to receive the events.

```mermaid
graph TD
    subgraph NetworkPeerActor
        NPA[NetworkPeerActor]
    end

    subgraph DialerActors
        DA1[DialerActor 1]
        DA2[DialerActor 2]
        DA3[DialerActor 3]
    end

    subgraph EventBus
        EB[EventBus]
    end

    %% Connections
    NPA --> EB
    EB --> DA1
    EB --> DA2
    EB --> DA3
```

packages/ciphernode/net/src/network_peer.rs
```diff
@@ -29,135 +53,302 @@ pub struct NodeBehaviour {
...
+        quic_port: Option<u16>,
+        enable_mdns: bool,
+    ) -> Addr<Self> {
+        let peer = Self::new(id, initial_peers, quic_port, enable_mdns).start();
+        peer
     }
 
-    pub fn tx(&self) -> mpsc::Sender<NetworkPeerCommand> {
-        self.cmd_tx.clone()
+    fn begin_swarm_polling(&mut self, ctx: &mut Context<Self>) {
```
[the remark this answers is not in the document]
ryardley: I see from the comment above that we need this for managing the swarm stream. If that is the case then we should look carefully for other simpler ways to handle this.

packages/ciphernode/net/src/bin/p2p_test.rs
```diff
@@ -1,132 +1,161 @@
...
+}
+
+impl Actor for TestManager {
+    type Context = Context<Self>;
+
+    fn started(&mut self, ctx: &mut Self::Context) {
+        info!("TestManager for '{}' started", self.name);
 
-#[tokio::main]
+        ctx.run_interval(Duration::from_millis(100), |act, _ctx| {
```
ryardley, line 39: This might also be simplified with an EventBus

deploy/agg.yaml
```diff
@@ -5,7 +5,6 @@ peers:
   - "/dns4/cn1/udp/9091/quic-v1"
   - "/dns4/cn2/udp/9092/quic-v1"
   - "/dns4/cn3/udp/9093/quic-v1"
-  - "/dns4/aggregator/udp/9094/quic-v1"
```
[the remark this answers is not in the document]
hmzakhalid: I thought that was a mistake as i was seeing failed attempts to itself. so i removed them.

packages/ciphernode/net/src/network_peer.rs
hmzakhalid: Yeah the eventbus would simplify a lot of this

packages/ciphernode/net/src/network_peer.rs
hmzakhalid: To break it down simply. We **can't** just spawn a continuous async stream on the Swarm (e.g. `ctx.spawn(async move { loop { swarm.select_next_some().await } })`) because borrowing `self.swarm` across `.await` inside the actor violates Rust’s lifetime rules. So by polling in a short interval or a run-later loop, we can keep the Swarm’s ownership in the actor and never hold a `&mut self` borrow across an `.await`. That avoids having to wrap the Swarm in `Arc<Mutex>` or moving it out of the actor entirely. essentially manual polling is a simple workaround that preserves the actor’s control over the Swarm and satisfies Rust’s ownership constraints.

packages/ciphernode/net/src/network_peer.rs
```diff
@@ -29,135 +53,302 @@ pub struct NodeBehaviour {
...
-    /// The gossipsub topic that the peer should listen on
-    topic: gossipsub::IdentTopic,
-    /// Broadcast channel to report NetworkPeerEvents to listeners
-    event_tx: broadcast::Sender<NetworkPeerEvent>,
-    /// Transmission channel to send NetworkPeerCommands to the NetworkPeer
-    cmd_tx: mpsc::Sender<NetworkPeerCommand>,
-    /// Local receiver to process NetworkPeerCommands from
-    cmd_rx: mpsc::Receiver<NetworkPeerCommand>,
+    quic_port: Option<u16>,
+    dialer: Option<Addr<DialerActor>>,
```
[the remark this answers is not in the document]
hmzakhalid: Agreed.

ryardley:
After our discussion we have found that in order to move forward with this it makes a lot of sense to look into replacing Actix before doing this.

hmzakhalid:
@ryardley This is ready for review now!

packages/ciphernode/events/src/enclave_event/mod.rs
```diff
@@ -139,6 +139,11 @@ impl EnclaveEvent {
             _ => false,
         }
     }
+
+    pub fn event_type(&self) -> String {
```
ryardley, line 143: This looks duplicated below so I believe we can delete this.

packages/ciphernode/net/src/bin/p2p_test.rs
```diff
@@ -78,26 +93,28 @@ async fn main() -> Result<()> {
...
+                        );
+                        received.insert(String::from_utf8(msg).unwrap());
                     }
-                    Err(e) => println!("{} received invalid UTF8: {}", name, e),
-                },
-                _ => (),
+                    _ => (),
+                }
             }
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
ryardley, line 117: Is this necessary?

packages/ciphernode/net/src/dialer.rs
```diff
@@ -1,210 +1,261 @@
...
-    let opts: DialOpts = multi.clone().into();
-    let dial_connection = opts.connection_id();
-    info!("Dialing: '{}' with connection '{}'", multi, dial_connection);
-    cmd_tx
-        .send(NetworkPeerCommand::Dial(opts))
-        .await
-        .map_err(to_retry)?;
-    wait_for_connection(&mut event_rx, dial_connection).await
+#[derive(Clone)]
+pub struct DialerActor {
```
ryardley, line 30: Could call this `Dialer`? just we haven't used the `Actor` prefix before.

packages/ciphernode/events/src/enclave_event/mod.rs
hmzakhalid: Good Catch!

packages/ciphernode/net/src/dialer.rs
```diff
@@ -1,210 +1,261 @@
...
+            listener: addr.clone().recipient(),
+        });
+
+        addr
+    }
+
+    pub fn dial_peer(
+        addr: String,
+        net_bus: Addr<EventBus<NetworkPeerEvent>>,
+        tx: mpsc::Sender<NetworkPeerCommand>,
```
ryardley, line 60: Why do we need a `tx` here? Why can we not simply use Actor messages? We probably need: 

```
enum NetworkMessage {
  Event(NetworkPeerEvent),
  Command(NetworkPeerCommand)
}
```

or something similar.

packages/ciphernode/net/src/dialer.rs
hmzakhalid: Yup!

packages/ciphernode/net/src/dialer.rs
hmzakhalid: They don't play well as the trait `Clone` is not implemented for `DialOpts` inside `NetworkPeerCommand`.
Which is required by the `Event` Trait and is not removeable. 
So the easiest solution was to use a mpsc tx

packages/ciphernode/net/src/dialer.rs
```diff
@@ -1,210 +1,261 @@
...
-    cmd_tx
-        .send(NetworkPeerCommand::Dial(opts))
-        .await
-        .map_err(to_retry)?;
-    wait_for_connection(&mut event_rx, dial_connection).await
+#[derive(Clone)]
+pub struct DialerActor {
+    net_bus: Addr<EventBus<NetworkPeerEvent>>,
+    tx: mpsc::Sender<NetworkPeerCommand>,
+    pending_connection: HashMap<ConnectionId, PendingConnection>,
```
ryardley, line 33: Assuming we are spawning a new Dialer per address connection. Do we need this to be a HashMap? Could it simply be `Option<PendingConnection>`? We could trigger the first dial attempt in the `started()` handler. Then when we create a `Dialer` actor it will autonomously dial a connection until it fails or succeeds.

packages/ciphernode/net/src/dialer.rs
ryardley: Can we avoid sending DialOpts in an event and instead send something serializable?

ryardley: Might be able to make it send `Arc<DialOpts>` instead?

ryardley: Also can use `CorrelationId` instead of `ConnectionId` from `DialOpts`

ryardley: Yeah sending `Arc<DialOpts>` works

packages/ciphernode/net/src/dialer.rs
hmzakhalid: Yeah thats much better, refactored it

packages/ciphernode/net/src/dialer.rs
hmzakhalid: We could do that by moving the address resolution logic to the network peer instead of the dialer. 
I guess thats a small cost to pay for a whole lot of simplification.

hmzakhalid: > Yeah sending `Arc<DialOpts>` works

Can't clone it in network peer `self.swarm.dial((*dial_opts).clone())`

ryardley (commented):
1. I think we should re-consider splitting up `NetworkPeer` into "Actor" and "Inner" and use the "Actor" to do things like correlations on synchronous things like dialing and errors. We can then use the retry function we already have and call it within the `Dialer` actor to handle the retry functionality in a really simple and reusable way. 
2. We should only use channels to communicate between libp2p struct and the wrapper actor and should favor a unified event for the event bus that is both commands and events. (Splitting them up is as confusing as using separate channels)

hmzakhalid:
> 1. I think we should re-consider splitting up `NetworkPeer` into "Actor" and "Inner" and use the "Actor" to do things like correlations on synchronous things like dialing and errors. We can then use the retry function we already have and call it within the `Dialer` actor to handle the retry functionality in a really simple and reusable way.
> 2. We should only use channels to communicate between libp2p struct and the wrapper actor and should favor a unified event for the event bus that is both commands and events. (Splitting them up is as confusing as using separate channels)

There's an issue with `NetworkPeer` and `NetworkPeerInner` split.
The NetworkPeerInner cannot be stored directly in the NetworkPeer actor because Swarm cannot be cloned.

Which has some important implications:

1. We can't directly call methods on NetworkPeerInner from NetworkPeer because:
   - The Inner must live in its own task/thread
   - We can't share or clone the Swarm between tasks

2. This forces us to use channels for all communication:
   - mpsc channel for sending commands to the Inner
   - broadcast channel for receiving SwarmEvents from the Inner

and on the broadcast channel suggestion - even with a broadcast channel, we still need an mpsc channel for commands because:
   - Any interaction with the Swarm (like dialing) must go through its own task
   - We can't directly call methods on the Swarm from the NetworkPeer actor

The architecture would look like:
NetworkPeer (Actor) -> mpsc -> NetworkPeerInner (SwarmOperations (Dial, Publish))
NetworkPeerInner -> broadcast -> NetworkPeer (SwarmEvents)

ryardley:
nACK -  we were not arriving at a good architecture due to unnecessary complexity and I think if we were to implement this we would possibly revert it. This was a spike and I feel like it is not worth working on at this point until we clearly see the need for it. I suggest this should be closed.

hmzakhalid:
Agreed, Closing this. Doesn’t seem worth pursuing further at this stage.

> nACK - we were not arriving at a good architecture due to unnecessary complexity and I think if we were to implement this we would possibly revert it. This was a spike and I feel like it is not worth working on at this point until we clearly see the need for it. I suggest this should be closed.

Issue #240, opened 2025-01-24
Implement Excubiae for verification


ryardley:
Excubiae usecase:
- **Aggregating and combining checkers** within Enclave (Greco PK Encryption circuit, Allowed Data Source Proof (Semaphore), Data Validation Proof) - aggregation functionality is [not yet available](https://github.com/privacy-scaling-explorations/excubiae/issues/13) however we can simply compose our own aggregator based on a strategy pattern
- **Reuse existing deployments** Using preexisting deployed checkers within new policies - we wont need this initially however once we have deployed some checkers for specific circuits this may be useful.

We might want to apply Excubiae for the following reasons:
- Connect to PSE developer community
- Potentially allow for Enclave verifications to be used and composed within external protocols in ways we have not yet thought about

```mermaid
classDiagram
    class IChecker {
        <<interface>>
    }
    
    class IBaseChecker {
        <<interface>>
        +check(subject: address, evidence: bytes[]): bool
    }
    
    class IPolicy {
        <<interface>>
    }
    
    class IBasePolicy {
        <<interface>>
        +enforce(subject: address, evidence: bytes[]): void
    }
    
    class CorrectEncryptionChecker {
        +check(subject: address, evidence: bytes[]): bool
    }
    
    class AllowedDataSourceChecker {
        +check(subject: address, evidence: bytes[]): bool
    }
    
    class DataValidationChecker {
        +check(subject: address, evidence: bytes[]): bool
    }
    
    class InputValidationChecker {
        -encryptionChecker: CorrectEncryptionChecker
        -dataSourceChecker: AllowedDataSourceChecker
        -dataValidationChecker: DataValidationChecker
        +check(subject: address, evidence: bytes[]): bool
    }
    
    class InputValidationPolicy {
        -validationChecker: InputValidationChecker
        +enforce(subject: address, evidence: bytes[]): void
    }
    
    IChecker <|-- IBaseChecker
    IPolicy <|-- IBasePolicy
    IBaseChecker <|.. CorrectEncryptionChecker
    IBaseChecker <|.. AllowedDataSourceChecker
    IBaseChecker <|.. DataValidationChecker
    IBaseChecker <|.. InputValidationChecker
    IBasePolicy <|.. InputValidationPolicy
    
    InputValidationChecker o-- CorrectEncryptionChecker
    InputValidationChecker o-- AllowedDataSourceChecker
    InputValidationChecker o-- DataValidationChecker
    
[...]

PR #241, opened 2025-01-27
Excubiae Integration

ryardley:
Present a version of utilizing Excubiae for managing input validation.

1. User makes a request specifying an inputlimit variable for the request
2. As the E3Program validates that request it deploys a policy (proxy) for handing the request
3. The policy proxy is used to manage state for the request as the address of the policy is stored on the E3 instance and thus is scoped to the e3id

At a later date we can do things like work out how to aggregate checkers especially when Excubiae creates the aggregation API which is yet to be specified. 

## Rhetorical considerations for reviewers

- is the complexity here worth the benefit. I am still on the fence as I believe we could come up with a simpler but possibly less flexible way to do this - however this also works and I don't really have a real issue with it despite contracts like `MockInputValidatorPolicyFactory` - I guess we just need to be sure this is easily auditable.
- The aggregation API has not yet been revealed. What if it is more and more complex? It is probably fine but there might be further risks here.  

closes: #240 

---

- [x]  Workout how to best connect to the excubiae contracts. This will likely require automation for updating contracts locally.
- [x] Create `MockInputValidatorChecker` 
- [x] Create `InputValidatorPolicy` which uses `IBaseChecker`
- [x] Swap `IInputValidator` for `IBasePolicy`
- [x] Call `.enforce(address, evidence)` instead of `.validate` and update to use `InputValidatorPolicy`.
- [x] Remove `MockInputValidator`
- [x] Fix up failing merkletree test as this requires conversion from BasePolicy to AdvancedPolicy
- [x] Test that inputLimit is configurable
- [x] Add test to check that subject can submit to a second E3 request when the policy is limited to a single enforcement point per E3Request
- [x] Remove update script in favor of npm as it looks like they are deploying regularly now and update with latest changes from upstream: https://github.com/privacy-scaling-explorations/excubiae/commits/main/

## Summary by CodeRabbit

- **New Features**
  - Introduced an input limit parameter for requests, allowing control over how many inputs can be submitted.
  - Enhanced input validation and policy management with improved enforcement and error feedback.
  - Added event notifications that signal key policy actions.
  - New interfaces and contracts for managing enclave policies and validation processes.
  - Added a new contract for deploying enclave policy factories.

- **Tests**
  - Updated test scenarios to validate the new input limit functionality and the refined validation processes.
  - Introduced new fixtures for easier testing of the input validator checker and policy factory.

- **Chores**
  - Upgraded external dependencies and streamlined deployment scripts for improved integration and maintainability.


ryardley:
So after doing this it occurred to me that a flaw in this implementation is that we are checking subject -> target -> state and that the subject and target here must be an address. This means we kind of need one address per E3Request. We may need to somehow generate an address from the e3_id which might be a possible workaround.

ryardley:
> So after doing this it occurred to me that a flaw in this implementation is that we are checking subject -> target -> state and that the subject and target here must be an address. This means we kind of need one address per E3Request. We may need to somehow generate an address from the e3_id which might be a possible workaround.

The new version of Excubiae means this will be a new proxy contract deployed per E3Request via the [Factory](https://github.com/privacy-scaling-explorations/excubiae/blob/main/packages/contracts/contracts/src/core/proxy/Factory.sol)

packages/evm/scripts/update_excubiae.sh
```diff
@@ -0,0 +1,77 @@
...
+#
+# 1. Performing a sparse checkout of the latest contracts
+# 2. Copying them to our local contracts directory
+#
+# Why This Approach?
+#
+# We chose this method because:
+#
+# * The NPM package is outdated and doesn't reflect the latest changes
+# * Since we use Hardhat instead of Foundry, we can't use Foundry's library management
```
auryn-macmillan, line 16: We could migrate to foundry if this would make life easier.
I have a slight preference for HH, purely because it's what I'm most comfortable in/familiar with, but happy to adapt.

ryardley: They released a new version with the refactor which makes me feel confident they will update the releases with a cadence we can follow instead of working off the master branch. My plan at this stage is to move to npm and follow their release schedule.

ryardley:
- [ ] fix integration test to use correct request api
- [ ] investigate coderabbit suggestions

auryn-macmillan:
Tagging @0xjei here as some of this might be good feedback for Excubiae.

0xjei:
> Tagging @0xjei here as some of this might be good feedback for Excubiae.

thanks @auryn-macmillan! excited to see Enclave exploring excubiae!!! Today, I have released the [v0.3.1](https://github.com/privacy-scaling-explorations/excubiae/releases/tag/v0.3.1) and we now support extensions (see [Semaphore](https://github.com/privacy-scaling-explorations/excubiae/tree/main/packages/contracts/contracts/extensions) derived from the [MACI Gatekeeper](https://github.com/privacy-scaling-explorations/maci/blob/dev/packages/contracts/contracts/gatekeepers/SemaphoreGatekeeper.sol)) and more. The [technical document](https://hackmd.io/@0xjei/B1RXoTh71e) is up-to-date but feel free to ask any questions / ping me whenever you need anything or write in our discord channel #excubiae (PSE discord).

ryardley:
- [ ] Update to `0.3.1`

0xjei:
> * [ ]  Update to `0.3.1`

do you need any help to update to v0.3.1?

ryardley:
It is on my todo list to look at - have been looking at other stuff last little while. Happy for some help or a suggestion if that is easy for you.

packages/evm/contracts/test/MockInputValidatorChecker.sol
```diff
@@ -0,0 +1,30 @@
...
+
+/// @title MockInputValidatorChecker.
+/// @notice Enclave Input Validator
+/// @dev Extends BaseChecker for input verification.
+contract MockInputValidatorChecker is AdvancedChecker {
+    /// @notice Validates input
+    /// @param subject Address to check.
+    /// @param evidence mock proof
+    /// @return True if proof is valid
+    function _checkMain(
```
0xjei, line 16: nit: do we want to add `_checkPre` and `_checkPost` mocks to reach a more complete code coverage?

packages/evm/contracts/test/MockInputValidatorPolicy.sol
```diff
@@ -0,0 +1,53 @@
...
+            .decode(data, (address, address, uint8));
+        _transferOwnership(sender);
+
+        ADVANCED_CHECKER = AdvancedChecker(advCheckerAddr);
+        SKIP_PRE = true;
+        SKIP_POST = true;
+        inputLimit = _inputLimit;
+    }
+
+    function enforceWithLimit(
```
0xjei, line 34: nit: maybe is better to maintain consistency in naming with `enforce` since we are using `super._enforce()`.

packages/evm/contracts/test/MockInputValidatorPolicy.sol
```diff
@@ -0,0 +1,53 @@
...
+import { Check } from "@excubiae/contracts/interfaces/IAdvancedChecker.sol";
+
+/// @title BaseERC721Policy.
+/// @notice Policy enforcer for Enclave Input validation.
+/// @dev Extends BasePolicy with Enclave specific checks.
+contract MockInputValidatorPolicy is AdvancedPolicy, IEnclavePolicy {
+    error MainCalledTooManyTimes();
+
+    uint8 public inputLimit;
+    mapping(address subject => uint8 count) public enforced;
```
0xjei, line 18: +1 this is the correct way to keep tracking of the enforcements.

packages/evm/test/Enclave.spec.ts
```diff
@@ -541,24 +594,25 @@ describe("Enclave", function () {
...
-          request.startTime,
-          request.duration,
-          request.e3Program,
-          request.e3ProgramParams,
-          request.computeProviderParams,
-          { value: 10 },
-        ),
-      ).to.be.revertedWithCustomError(enclave, "InvalidComputationRequest");
-    });
+    // XXX: work out a different way to test this
```
0xjei, line 597: nit: maybe add a //todo comment here?

packages/evm/test/Enclave.spec.ts
```diff
@@ -36,19 +37,61 @@ const proof = "0x1337";
...
+    const { inputValidatorChecker, inputValidatorPolicyFactory } =
+      await deployInputValidatorContracts();
+
     const e3Program = await deployE3ProgramFixture(
-      await inputValidator.getAddress(),
+      await inputValidatorPolicyFactory.getAddress(),
+      await inputValidatorChecker.getAddress(),
     );
 
+    // TODO: is this too restrictive?
```
0xjei, line 90: note: can you expand this todo?

packages/evm/package.json
```diff
@@ -85,10 +85,13 @@
     "prettier:write": "prettier --write \"**/*.{js,json,md,sol,ts,yml}\"",
     "task:deployLock": "hardhat task:deployLock",
     "task:withdraw": "hardhat task:withdraw",
+    "update:excubiae": "./scripts/update_excubiae.sh",
```
0xjei, line 88: I cannot see any `scripts` folder. Where is it located? Is it autogenerated?

0xjei (dismissed):
uTACK. 

@ryardley you did an amazing work and understood Excubiae correctly. I just left a few nits and notes. Feel free to tackle them or merge directly, lg in either cases to me. Thanks!

packages/evm/contracts/test/MockInputValidatorChecker.sol
ryardley: I am not sure we need this usecase yet as we don't quite know the shape of all the validations etc. we need to do. My feeling is that we can roll with this and refactor as we cement what an input validation looks like.

packages/evm/contracts/test/MockInputValidatorPolicy.sol
ryardley: This lead to errors as `AdvancedPolicy` which we are inheriting from doesn't have a virtual function.

packages/evm/test/Enclave.spec.ts
ryardley: I added some more context around what I was thinking.

packages/evm/test/Enclave.spec.ts
ryardley: I have resolved this and reinstated the test.

packages/evm/contracts/test/MockInputValidatorChecker.sol
ryardley: Yeah I know but I needed the advanced policy to allow for multiple main checks.

packages/evm/contracts/test/MockInputValidatorPolicy.sol
0xjei: yeah, you just need to override the `_enforce()` without caring much. The `AdvancedPolicy` will take care of calling it from is `enforce()` external function! 

See [SemaphoreExcubiae](https://github.com/privacy-scaling-explorations/excubiae/blob/9bf739a43932051fa3aca4daac0b195deaad5d7b/packages/contracts/contracts/extensions/SemaphorePolicy.sol#L28) as example.

ryardley: This worked great thanks!

0xjei (approved):
uTACK

PR #339, opened 2025-04-14
🧟 Fix Zombie Process

### LINKED ISSUE #316: 🧟 Zombie Process! Process is not terminating
We should be able to initiate a shutdown with: `pkill -15 enclave`

This returns the terminal but doesn't shutdown.

At the end of our integration tests we have some errors when the binaries fail to clean up an open websocket. This prevents them from shutting down gracefully and should be investigated. 

Image

We may want to refactor handles so we get a handle to every async process including and especially the evm_reader and join them.

This should incorporate #162 

---

### To Test

To test alter https://github.com/gnosisguild/enclave/blob/main/tests/integration/base.sh by adding an infinite sleep after launching the nodes:

Image

Then run:

```
pnpm test:integration base
```

Wait for the sleep.

In a second terminal:

```
ps aux | grep enclave
```

You should see the running enclave processes:

Image

Now run in that terminal:

```
pkill -15 -f enclave
```

You will get lots of output detailing a shutdown process:

Image

However there are still running enclave processes:

Image

We can only shut these processes down by running:

```
pkill -9 -f enclave
```

ryardley:
Closes #316 

## Testing

Easiest way to test this is to run the integration tests locally and ensure we don't get errors at the end of the script:

```
pnpm test:integration base
```

This is what the test is doing on `main`:

Image

You can also look at following the test instructions in the #316 issue too. If you want a more focused test.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced asynchronous shutdown handling to ensure proper error propagation and prevent indefinite waiting.
  - Improved event stream cleanup by adding robust unsubscription logic with clear success or error feedback.

- **Tests**
  - Refined integration test scripts by adjusting wait durations and process termination commands for more reliable and efficient testing.
  - Added new functions to better manage process counts and graceful shutdowns in integration tests.

- **Chores**
  - Removed unused code to streamline operations and simplify maintenance.
  - Introduced a new shutdown command for improved process termination.

### DIFF
--- a/packages/ciphernode/enclave/src/aggregator_start.rs
+++ b/packages/ciphernode/enclave/src/aggregator_start.rs
@@ -17,9 +17,6 @@ pub async fn execute(
         aggregator_start::execute(config, pubkey_write_path, plaintext_write_path).await?;
 
     info!("LAUNCHING AGGREGATOR {}", peer_id);
-    tokio::spawn(listen_for_shutdown(bus.into(), handle));
-
-    std::future::pending::<()>().await;
-
+    tokio::spawn(listen_for_shutdown(bus.into(), handle)).await?;
     Ok(())
 }
--- a/tests/integration/fns.sh
+++ b/tests/integration/fns.sh
@@ -178,6 +178,7 @@ gracefull_shutdown() {
   kill_em_all
 }
 
+# Run this at the start of every test to ensure we start with a clean slate
 kill_em_all
 
 # Set up trap to catch errors and interrupts
--- a/tests/integration/fns.sh
+++ b/tests/integration/fns.sh
@@ -43,8 +43,7 @@ ENCLAVE_BIN=./packages/ciphernode/target/debug/enclave
 # Function to clean up background processes
 cleanup() {
     echo "Cleaning up processes..."
-    jobs -p
-    [ -n "$(jobs -p)" ] && kill -9 $(jobs -p) 2>/dev/null || true
+    jobs -p | xargs -r kill -9 2>/dev/null || true
     exit ${1:-1}
 }
 
--- a/tests/integration/fns.sh
+++ b/tests/integration/fns.sh
@@ -150,10 +150,9 @@ kill_proc() {
   kill $pid
 }
 
-metalica() {
-  # kill_em_all
+kill_em_all() {
   pkill -9 -f "target/debug/enclave" || true
-  pkill -9 -f "hardhat node" || true
+  pkill -9 -f "hardhat" || true
 }
 
 launch_evm() {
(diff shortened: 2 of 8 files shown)


ryardley:
**Question**: Do we need a test specifically for this to catch regression?

ryardley:
Ok so now if the binary has not completed a graceful shutdown after 10 seconds the test should fail.

tests/integration/fns.sh
```diff
@@ -163,7 +163,22 @@ launch_evm() {
   fi
 }
 
-metallica
+ensure_process_count_equals() {
+  local process_name="$1"
+  local expected_count="$2"
+  local actual_count=$(ps aux | grep "$process_name" | grep -v "grep" | grep -v $$ | wc -l)
+  [ "$actual_count" -eq "$expected_count" ]
+  return $?
```
hmzakhalid, line 171: This can still produce false positives. we should use `pgrep`, which was designed for this exact purpose. For instance:
> 
> ```bash
> local actual_count="$(pgrep -f "$process_name" | wc -l)"
> [ "$actual_count" -eq "$expected_count" ]
> return $?
> ```
> 
This avoids having to manually strip out `grep` processes and is less prone to accidental matches.

hmzakhalid:
tACK - kill em all

tests/integration/fns.sh
ryardley: Can also get false positives with pgrep as I actually was - which is switched to this as it is easier to debug!

packages/ciphernode/enclave/src/aggregator_start.rs
```diff
@@ -17,9 +17,6 @@ pub async fn execute(
         aggregator_start::execute(config, pubkey_write_path, plaintext_write_path).await?;
 
     info!("LAUNCHING AGGREGATOR {}", peer_id);
-    tokio::spawn(listen_for_shutdown(bus.into(), handle));
-
-    std::future::pending::<()>().await;
-
+    tokio::spawn(listen_for_shutdown(bus.into(), handle)).await?;
```
0xjei, line 20: +1 very neat

tests/integration/fns.sh
```diff
@@ -163,7 +163,22 @@ launch_evm() {
...
+}
+
+gracefull_shutdown() {
+  pkill -15 -f "target/debug/enclave" || true
+  sleep 10
+  ensure_process_count_equals "target/debug/enclave" 0 || return 1
+  kill_em_all
+}
+
+kill_em_all
```
0xjei, line 181: is this duplicated since it's called on `gracefull_shutdown()` or?

0xjei (commented):
uTACK - left some comments

tests/integration/fns.sh
hmzakhalid: Alrighty!

tests/integration/fns.sh
ryardley: No this runs before every test. Will add a comment.

0xjei: makes sense, thx for the quick explanation!

0xjei (approved):

PR #345, opened 2025-04-16
Enforce valid parameters

### LINKED ISSUE #269: Valid Parameters
When a computation is requested, the chosen encryption scheme parameters (e.g., BFV) must be proven valid.

A simple initial solution would be to only allow users to select from one or several hard-coded sets of valid parameters.

0xjei:
This PR aims to provide a basic approach to validating a specific set of E3 program parameters for the selected FHE scheme (e.g. BFV) to be one of the potential stored on-chain. For (enc)(dec)oding, we will use the standard ABI encoder, which is a cross-library, cvm-friendly standard. Therefore, we will use `alloy.rs` on the Rust side while using `ethers.js` on the TS side without adding any new crates/packages.

Notable changes:
- `evm`: added a mapping to `enclave.sol` to store the encoded set of parameters (`bytes`) as boolean. If `true`, the set can be selected, otherwise a custom error is thrown. The valid set of params is provided as `bytes[]` at `initialise()` time (aka deploy). To support these changes, the `fixture' for tests and `tasks' for deployments have been modified accordingly.
- `ciphernodes` && `CRISP`: Updated methods to support encoding with ABI.

Notes:
This approach is flexible enough to support multiple schemes (e.g. BFV, TFHE) with minimal changes. For now, we will only mirror the parameters currently required for `BFV'.

--

closes #269

## Summary by CodeRabbit

- **New Features**
  - Introduced support for registering and managing allowed E3 program parameters in the Enclave contract, including a new public mapping, event emission, and owner-only configuration function.
  - Deployment scripts and test fixtures updated to support passing encoded E3 program parameters during contract deployment.

- **Bug Fixes**
  - Improved handling and validation of E3 program parameters in contract interfaces and mock contracts.

- **Refactor**
  - Streamlined serialization and ABI encoding/decoding of BFV parameters for improved clarity and reliability.
  - Simplified parameter construction and decoding logic in cryptographic modules.

- **Documentation**
  - Enhanced inline documentation for contract interfaces and struct fields.

- **Tests**
  - Added comprehensive tests for E3 program parameter management and validation in the Enclave contract.
  - Updated test cases to align with new encoding and validation logic.

- **Chores**
  - Added and reorganized dependencies in manifest files for improved maintainability.
  - Minor code style and formatting improvements across several modules.


0xjei:
hello @ryardley @hmzakhalid - I tried to minimise the changes and impacts here across all the stack. I tested the CLI and everything else and should be fine. Open to any feedbacks / comments since it's the first I'm working on across the stack.

As a side note, I think there's lots of margin for improvement and code modularisation. However, the goal is to make things work now - we can iterate back later.

ryardley:
Hey this looks like a great start here and makes sense. I have scanned this and will look over it in more detail tomorrow and next couple of days.

The two things that jump out to me that I think are really important we should see in this PR would be:

1. Setup the ability to import our encode and decode lib from a shared library and removing the duplication.
2. Some simple unit tests on the encoding and decoding.

Setting up a shared Cargo workspace at the root is probably really useful (although not technically mandatory) as it would also allow us to run cargo commands at the root of the repo.

packages/ciphernode/fhe/src/ext.rs
```diff
@@ -41,7 +42,9 @@ impl E3Extension for FheExtension {
             ..
         } = data.clone();
 
-        let Ok(fhe_inner) = Fhe::from_encoded(&params, seed, self.rng.clone()) else {
+        let params_bytes: Vec<u8> = params.iter().map(|&x| x as u8).collect();
```
ryardley, line 45: is params here already a `Vec<u8>`?

examples/CRISP/apps/server/src/server/utils.rs
```diff
@@ -1,24 +1,78 @@
+use alloy::dyn_abi::{DynSolType, DynSolValue};
```
ryardley, line 1: Is there a reason for not using the static encoder?

https://docs.rs/alloy-sol-types/latest/alloy_sol_types/

packages/ciphernode/events/src/enclave_event/e3_requested.rs
```diff
@@ -17,8 +17,8 @@ impl Display for E3Requested {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "e3_id: {}, threshold_m: {}, src_chain_id: {}, seed: {}, params: <omitted>",
-            self.e3_id, self.threshold_m, self.src_chain_id, self.seed
+            "e3_id: {}, threshold_m: {}, src_chain_id: {}, seed: {}, params: {:?}",
+            self.e3_id, self.threshold_m, self.src_chain_id, self.seed, self.params
```
ryardley, line 21: Just fyi this was deliberate and done in order to avoid spamming the console with huge data blobs but it might be good to work out a better way to do it.

packages/evm/deploy/enclave.ts
```diff
@@ -36,10 +37,19 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
   // Deploy Enclave contract
+  const polynomial_degree = ethers.toBigInt(2048);
+  const plaintext_modulus = ethers.toBigInt(1032193);
+  const moduli = [ethers.toBigInt("4503599626321921")];
+
+  // Encode just the struct (NOT the function selector)
+  const encoded = ethers.AbiCoder.defaultAbiCoder().encode(
+    ["uint256", "uint256", "uint256[]"],
+    [polynomial_degree, plaintext_modulus, moduli],
+  );
```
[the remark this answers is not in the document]
0xjei: to be done when refactoring - link new issue here.

packages/ciphernode/fhe/src/fhe.rs
```diff
@@ -43,12 +44,66 @@ impl Fhe {
...
+                    .collect::<Result<Vec<u64>>>()?
+            } else {
+                return Err(anyhow!("Expected Array for moduli"));
+            };
+
+            Ok((degree, plaintext_modulus, moduli))
+        } else {
+            Err(anyhow!("Expected Tuple for decoded result"))
+        }
+    }
```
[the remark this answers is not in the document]
0xjei: same as above

0xjei:
@ryardley okay, got it.

examples/CRISP/apps/server/src/server/utils.rs
0xjei: yes, `sol!` it wraps values in a tuple, adding a 32-byte offset at the beginning of the encoding.

0xjei:
seems pretty hard to reconcile the two branches. maybe is preferable to avoid inconsistency and waste of time and just move the key changes inside the bfv #388 in favour of this!

0xjei:
this is almost ready - I just need to adapt CRISP `apps` and should be ready for review. I will complete it tomorrow.

0xjei:
encountering a small difficulty but this is due to the fact that the chain expect `alloy_primitives::Bytes` and I'm using `DynSolValue` for the abi encoder. Thinking about it.

0xjei:
found potential bug on `E3Requested` params treated and expected as `Vec<u8>` instead of raw `bytes` leading to implicit conversions errors. Fixing and uniforming things...

0xjei:
unlocked but getting vote failed when voting from the UI

packages/commons/src/bfv/mod.rs
```diff
@@ -121,55 +139,70 @@ pub fn deserialize_bfv_params_arc(bytes: &[u8]) -> Arc<BfvParameters> {
...
+    let decoded = bytes_type
+        .abi_decode(bytes)
+        .expect("Failed to ABI decode bytes");
+
+    match decoded {
+        DynSolValue::Bytes(inner_bytes) => {
+            BfvParameters::try_deserialize(&inner_bytes).expect("Could not decode Bfv Params")
+        }
+        _ => panic!("Expected bytes value in ABI encoding"),
+    }
```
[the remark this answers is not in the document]
0xjei: it's fine to `panic!` for now! we will iterate back on errors

examples/CRISP/apps/server/src/server/routes/voting.rs
```diff
@@ -46,7 +46,9 @@ async fn broadcast_encrypted_vote(data: web::Json<EncryptedVote>) -> impl Respon
     let encoded_params = Bytes::from(params_value.abi_encode_params());
 
     // Broadcast vote to blockchain
-    let contract = EnclaveContract::new(CONFIG.enclave_address.clone()).await.unwrap();
+    let contract = EnclaveContract::new(CONFIG.enclave_address.clone())
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
0xjei: this is a change not mandatory and should not be included in this pr

examples/CRISP/apps/wasm-crypto/Cargo.lock
ryardley, line 1: Question: why do we have so many `Cargo.lock` files changing without `Cargo.toml` files changing?

ryardley: I guess cargo update might have been run.

ryardley (approved):
🚀

PR #355, opened 2025-04-22
Remove Input Limit from Enclave

hmzakhalid:


0xjei (approved):
uTACK - agree on moving the `inputLimit` as constant at deploy time. great work

PR #371, opened 2025-04-28
Add Semaphore Voter Registration to CRISP

hmzakhalid:


examples/CRISP/contracts/CRISPRisc0.sol
```diff
@@ -30,35 +40,66 @@
...
-        initialize(_enclave, _policy, _verifier);
-    }
+        require(address(_enclave) != address(0), EnclaveAddressZero());
+        require(address(_verifier) != address(0), VerifierAddressZero());
+        require(address(_semaphore) != address(0), SemaphoreAddressZero());
+        require(
+            address(_checkerFactory) != address(0),
+            InvalidCheckerFactory()
+        );
+        require(address(_policyFactory) != address(0), InvalidPolicyFactory());
```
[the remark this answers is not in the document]
hmzakhalid: watchu on about, it works

examples/CRISP/contracts/CRISPRisc0.sol
```diff
@@ -74,16 +115,28 @@
...
         );
         require(paramsHashes[e3Id] == bytes32(0), E3AlreadyInitialized());
-
         paramsHashes[e3Id] = keccak256(e3ProgramParams);
 
-        return (ENCRYPTION_SCHEME_ID, policy);
+        // Create a new group
+        uint256 groupId = semaphore.createGroup(address(this));
+        groupIds[e3Id] = groupId;
+
```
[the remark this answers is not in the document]
hmzakhalid: Well well well
```
    /// @dev See {SemaphoreGroups-_createGroup}.
    function createGroup(address admin) external returns (uint256);
```

examples/CRISP/contracts/CRISPPolicy.sol
```diff
@@ -2,51 +2,85 @@
...
-        SKIP_POST = true;
+        BASE_CHECKER = BaseChecker(baseCheckerAddr);
         inputLimit = _inputLimit;
     }
 
-    function _enforce(
+    /// @notice Validate the input and return the vote.
+    /// @param subject The subject to validate the policy on.
+    /// @param evidence Abi-encoded `ISemaphore.SemaphoreProof`.
+    function validate(
```
0xjei, line 37: hmm this could lead to unexpected errors since the `_enforce()` is already called by the [enforce()](https://github.com/privacy-scaling-explorations/excubiae/blob/main/packages/contracts/contracts/policy/BasePolicy.sol#L32) external fn and must be exposed / called only from there where all the logic is inherited from [BasePolicy](https://github.com/privacy-scaling-explorations/excubiae/blob/a7c8b6bc0ba998b13920a7e8044bf5fac22efe4b/packages/contracts/contracts/policy/BasePolicy.sol#L39) contract.

Please, double check this. If you need to return the vote, you can just store them or emit an event in the `internal` fn.

packages/evm/contracts/Enclave.sol
```diff
@@ -232,7 +228,7 @@ contract Enclave is IEnclave, OwnableUpgradeable {
             InputDeadlinePassed(e3Id, e3.expiration)
         );
 
-        e3.inputValidator.enforce(msg.sender, data, Check.MAIN);
+        e3.inputValidator.validate(msg.sender, data);
```
0xjei, line 231: that's exactly my concern here - I think you should call `enforce()` and not `validate()` (which should not exist as second-way to enforcing the checks) here.

0xjei (commented):
utACK - but I'm not sure about the `validate` and `enforce` pattern in this case since opens up to multiple ways to enforce the checks. Please, double check since it may require changes but I'm not sure. 

ps. didn't spent much time on the frontend and mocks - will go over them tomorrow. Take this review as general comment to think about what I wrote above and as a high-level approval of correct interpretation of Excubiae.

examples/CRISP/apps/client/src/context/voteManagement/VoteManagement.context.tsx
```diff
@@ -101,7 +111,23 @@ const VoteManagementProvider = ({ children }: VoteManagementProviderProps) => {
...
   }, [wasmLoading, enclaveLoading])
 
-  // Update user state when wallet connection changes
+  useEffect(() => {
+    if (semaphoreIdentity) {
+      return;
+    }
+
+    if (identityPrivateKey) {
+      let identity = Identity.import(identityPrivateKey);
```
Subhasish-Behera, line 120: Should this be wrapped in a try block?
if `identityPrivateKey` verifies if this is non empty. but it can still be invalid for import method.

examples/CRISP/apps/client/src/hooks/semaphore/useSemaphoreGroupManagement.ts
```diff
@@ -0,0 +1,132 @@
...
+            undefined,
+        query: {
+            enabled: groupId !== null && !!semaphoreIdentity,
+        },
+    });
+
+    const { data: fetchedMembers, isLoading: isLoadingMembers, refetch: refetchGroupMembers } = useReadContract({
+        address: E3_PROGRAM_ADDRESS,
+        abi: E3_PROGRAM_ABI,
+        functionName: 'getGroupCommitments',
```
Subhasish-Behera, line 52: I think the approach of generating the proof after getting the commiments/member from onchain data should be delegated to existing semaphore functions. That way we can be sure that there is no play around order of members registered, the code will be accurate whle supporting deregistration. 
the group required while generating proof can be done via,
```
            import { SemaphoreEthers } from "@semaphore-protocol/data";
            // Initialize SemaphoreEthers with the contract address
            const semaphoreEthers = new SemaphoreEthers("provider", {
                address: "semaphorecontract _address"
            });

            
            const members = await semaphoreEthers.getGroupMembers(groupId);
            
            const group = new Group(members);
```

hmzakhalid: that's how I implemented it [originally](https://github.com/gnosisguild/enclave/pull/371/commits/359dfdbe891cbc6ab98ed42e5f0f197b3143c2b3) using the semaphore/data lib. I removed it since I was storing the commitments for easier fetching, but I guess I overlooked handling updates or member removals. I'll revert the change.

examples/CRISP/contracts/CRISPPolicy.sol
hmzakhalid: I'm happy to sync anytime today to figure out a solution. The idea behind the `InputValidator` was to validate the proof and return the vote, that’s how it was originally set up. Since there might be multiple proof formats (e.g., some using Semaphore, some custom), we wanted to keep the input as generic bytes and let the validator handle decoding and validation. This led me to either introduce an intermediary contract (`InputValidator`) to handle `enforce`, or modify the policy to call `enforce` directly. I went with the latter to get things working quickly, we can now refine it together.

0xjei (commented):
utACK

making this as Comment instead of Approve since I focused my review mainly into contracts and excubiae integration more than other subsequent moving pieces. Kindly asking others to make a more serious review on other pieces.

ryardley:
I wouldn't mind understanding a bit more about the architectural direction here wrt excubiae/`IInputValidator` as we seem to be doing a reversion here. Do we need Excubiae at all in Enclave if we do this?

0xjei:
> I wouldn't mind understanding a bit more about the architectural direction here wrt excubiae/`IInputValidator` as we seem to be doing a reversion here and there is a shift in thinking. Do we need Excubiae at all in Enclave if we do this?

no, we do not need Excubiae in Enclave because means lock-in to it + overkill for some use cases. I proposed to @hmzakhalid as possibility to revert and have a simpler input validation. I think excubiae could be a huge advantage iff we reuse already deployed / developed Checker-Policy pair more than just creating a new one which is a small tweak of one already existent (e.g., Semaphore in this case). We were waiting for your input here, wdyt?

hmzakhalid:
> I wouldn't mind understanding a bit more about the architectural direction here wrt excubiae/`IInputValidator` as we seem to be doing a reversion here and there is a shift in thinking. Do we need Excubiae at all in Enclave if we do this? How do we plan to handle validating noir proof of valid encryption?

The core idea behind the `inputValidator` is that it should remain framework-agnostic. Enclave shouldn't directly depend on gatekeeping frameworks like Excubiae, instead, it's up to the application developer to decide how validations are handled. Whether they use Excubiae or another approach, their custom input validator is responsible for processing the input and returning validated data.

This keeps Enclave simple and gives developers full flexibility to implement validation logic as they see fit. All we require from them is the `E3ID` and raw data (as bytes). We pass the data to their validator, which can decode it however they choose (e.g., `abi.encode(SemaphoreProof, NiorProof, EncryptedVote)`, or other formats). Inside that validator, they can call Excubiae’s enforcement—or skip it entirely. The validator’s output is the clean, validated input.

This separation ensures that frameworks like Excubiae and Semaphore are implemented at the application level, not within Enclave. Some apps may have extensive validation; others might have none, it’s entirely up to the developer.

Happy to walk through the code tomorrow or keep things async, whichever works best.

packages/evm/contracts/interfaces/IEnclavePolicy.sol
```diff
@@ -3,10 +3,8 @@ pragma solidity >=0.8.27;
...
-import {
-    IAdvancedPolicy
-} from "@excubiae/contracts/interfaces/IAdvancedPolicy.sol";
+import { IBasePolicy } from "@excubiae/contracts/interfaces/IBasePolicy.sol";
 
 /// @title IEnclavePolicy.
-/// @notice Extends IPolicy with multi-phase validation capabilities.
-interface IEnclavePolicy is IAdvancedPolicy {}
+/// @notice Extends IPolicy with basic validation and enforcement capabilities.
+interface IEnclavePolicy is IBasePolicy {}
```
ryardley, line 10: Why do we still have this if we are removing Excubiae from Enclave?

packages/evm/contracts/interfaces/IEnclaveChecker.sol
```diff
@@ -3,10 +3,8 @@ pragma solidity >=0.8.27;
...
-import {
-    IAdvancedChecker
-} from "@excubiae/contracts/interfaces/IAdvancedChecker.sol";
+import { IBaseChecker } from "@excubiae/contracts/interfaces/IBaseChecker.sol";
 
 /// @title IEnclaveChecker.
-/// @notice Extends IAdvancedChecker
-interface IEnclaveChecker is IAdvancedChecker {}
+/// @notice Extends IBaseChecker
+interface IEnclaveChecker is IBaseChecker {}
```
ryardley, line 10: This too

0xjei:
> returning validated data

I'd like to add that this is not something excubiae is meant for: it just says "your data is fine for passing the following checks" - that's it. That's why we needed a wrapper to return the `vote` in the context of CRISP on top of excubiae. 

> This separation ensures that frameworks like Excubiae and Semaphore are implemented at the application level, not within Enclave. Some apps may have extensive validation; others might have none, it’s entirely up to the developer.

and this is also in the core logic of Excubiae. There's no advantage if you force it to some narrow protocol because nobody can generalize and reuse your already deployed Checker-Policy pair.

examples/CRISP/apps/client/src/context/voteManagement/VoteManagement.context.tsx
```diff
@@ -101,7 +111,23 @@ const VoteManagementProvider = ({ children }: VoteManagementProviderProps) => {
...
+      console.log('Semaphore identity loaded from storage.');
+      setSemaphoreIdentity(identity);
+    } else {
+      let identity = new Identity();
```
Subhasish-Behera: bit confused here. 
The idea was to create a deterministic Identity using the wallet to sign a message(as suggested here https://docs.semaphore.pse.dev/guides/identities#create-deterministic-identities see tips ) that stays constant during the round(ie roundID). But this seems like a generating a random identity. 
Plus I think the local storage of identity can be skipped by generating the identity again when generating proof(initially done when registering). u know the message, key from the wallet. so generating the identity shouldn't be a problem. I see the wallet connection as a requirement for every session, irrespective of castinga  vote or registration. Not sure, but I think generating identities again during the generation of proof is a more solid solution than trying to use stored identity as its more susceptible to changes specially in a browser scenario.

examples/CRISP/remappings.txt
```diff
@@ -5,3 +5,6 @@ risc0/=lib/risc0-ethereum/contracts/src/
 @gnosis-guild/enclave=node_modules/@gnosis-guild/enclave
 @excubiae/contracts=node_modules/@excubiae/contracts
 solady/=node_modules/solady
+@semaphore-protocol/contracts=node_modules/@semaphore-protocol/contracts
+@zk-kit/lean-imt.sol=node_modules/@zk-kit/lean-imt.sol
```
ryardley, line 9: These remappings need trailing slashes to work within the docker environement. Will post a PR to this PR as suggestion.

examples/CRISP/apps/client/src/context/voteManagement/VoteManagement.context.tsx
hmzakhalid: Good catch! Thanks for the feedback, I agree on deterministic identities and removing the local storage dependency. That's why the latest change generates the Identity using a combination of the wallet address and the round ID (new Identity `semaphore-identity-user.address-votingRound.round_id`. This makes it deterministic for the round without needing storage.
Regarding the use of message signing, while that's definitely more cryptographically robust way to link the wallet, I'm a bit concerned about the user experience. Adding an extra signature prompt every time the identity is needed introduces friction. For now, I've opted for the simpler deterministic approach.

0xjei (commented):
great progress @hmzakhalid - I have left some comments on the identity generation and nits.

examples/CRISP/apps/client/libs/wasm/pkg/package.json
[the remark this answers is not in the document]
hmzakhalid: yeah we need it in to generate proofs/encrypt vote, not sure how it got removed. I added it back after rudi notified me.

examples/CRISP/apps/client/src/context/voteManagement/VoteManagement.context.tsx
```diff
@@ -101,7 +114,23 @@ const VoteManagementProvider = ({ children }: VoteManagementProviderProps) => {
     setIsLoading(false)
   }, [wasmLoading, enclaveLoading])
 
-  // Update user state when wallet connection changes
+  useEffect(() => {
+    if (votingRound?.round_id && user?.address) {
+      const seedString = `semaphore-identity-${user.address}-${votingRound.round_id}`;
```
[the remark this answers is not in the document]
hmzakhalid: Thanks for pointing this out! and you're absolutely right that deterministic identities can be compromised. I had flagged this with @Subhasish-Behera earlier as well. For this example, I opted for the deterministic approach to avoid requiring users to sign and register for every round, which felt like a heavy UX tradeoff.
We could potentially explore combining the signing and registration into a single prompt to balance security and usability.

examples/CRISP/apps/client/src/context/voteManagement/VoteManagement.context.tsx
```diff
@@ -101,7 +114,23 @@ const VoteManagementProvider = ({ children }: VoteManagementProviderProps) => {
     setIsLoading(false)
   }, [wasmLoading, enclaveLoading])
 
-  // Update user state when wallet connection changes
+  useEffect(() => {
+    if (votingRound?.round_id && user?.address) {
```
0xjei, line 118:
```suggestion
    if (!(votingRound?.round_id == null) && user?.address) {
```

0xjei (changes requested):

examples/CRISP/apps/client/src/context/voteManagement/VoteManagement.context.tsx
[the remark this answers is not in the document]
ryardley: What I have seen before is you sign a known fixed message like "Signing this message will give this web application control of your CRISP account. Only sign this message if you are sure you trust the website you are connected to." with your ethers wallet this gives you a deterministic identity which can be recreated easily.

ryardley (approved):
Looks good to me 👽 🚀

Subhasish-Behera (approved):
reviewed the semaphore parts. LGTM!

PR #406, opened 2025-05-17
Revert "Enforce valid parameters (#345)"

ryardley:
This reverts commit 4997c1b621dd87aa2c28a27b2a8116b00ce6db96. It turns out this breaks the CRISP test so we should revert this merge the test and then debug this.

## Summary by CodeRabbit

- **Refactor**
  - Simplified serialization and deserialization of encryption parameters by switching to direct raw byte encoding and error handling.
  - Removed Solidity ABI encoding/decoding logic from both Rust and Solidity components.
  - Updated contract deployment scripts and tests to no longer require or handle encoded program parameters.
  - Improved import organization and minor formatting in several files.

- **New Features**
  - Enhanced error handling for parameter deserialization with clearer error contexts.

- **Breaking Changes**
  - Removed public contract functions and events related to setting and tracking allowed E3 program parameters.
  - Updated interface and contract function signatures to use calldata for byte parameters, improving efficiency.
  - Changed serialization/deserialization function signatures and removed some previously exported helper functions.

- **Tests**
  - Updated and simplified tests to align with new parameter handling and removed obsolete test cases.

- **Chores**
  - Updated dependencies by removing unused crates and adding necessary error handling libraries.

- **Documentation**
  - Removed or updated outdated comments related to parameter encoding and contract fields.

PR #407, opened 2025-05-17
Reapply "Enforce valid parameters (#345)" (#406)

ryardley:
Reapplying "Enforce valid parameters" this is the equivalent of #345 

This reverts commit 5c950eb667d07aef274d32a97246aaea6b39551e which reverted #345 

We need to merge the e2e test and then debug this to merge

### DIFF
--- a/examples/CRISP/apps/program/Cargo.lock
+++ b/examples/CRISP/apps/program/Cargo.lock
@@ -1391,7 +1391,7 @@ checksum = "67ba02a97a2bd10f4b59b25c7973101c79642302776489e030cd13cdab09ed15"
 name = "commons"
 version = "0.1.0"
 dependencies = [
- "anyhow",
+ "alloy-dyn-abi",
  "fhe",
  "fhe-traits",
 ]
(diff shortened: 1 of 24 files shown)


ryardley:
Looking at some of this stuff I am noticing there is a fair amount of lint noise - it would be good to solve that in a second PR also there are lots of random Cargo.lock files that shouldn't be part of this PR. One option that might be easier would be if we manually port the significant changes over to a new PR branching off `hacknet` to keep things clean? This might help also as the commons lib was refactored to be `enclave-sdk`.

0xjei:
> Looking at some of this stuff I am noticing there is a fair amount of lint noise - it would be good to solve that in a separate PR also there are lots of random Cargo.lock files that shouldn't be part of this PR. Not mandatory but one option here that might be easier would be if we manually port the significant changes over to a new PR branching off `hacknet` to keep things clean? This might help also as the commons lib was refactored to be `enclave-sdk`.

sweet, I can do that now.

closing based on comment above. will open a new accordingly and track here.

PR #412, opened 2025-05-18
PR Template wont work with github so reverting

ryardley:
After doing some experimentation on another repo it looks like we can't have multiple templates despite github suggesting you can with template query strings which would be awkward at best - so removing this seems the best course as we don't need a global template right now. 

## Summary by CodeRabbit

- **Chores**
  - Removed a GitHub pull request template used for standardizing PR submissions.


ryardley:
@coderabbit review

Issue #432, opened 2025-05-20
[Draft] Implement Observability setup for telemetry generated


Subhasish-Behera:
Parent Issue: https://github.com/gnosisguild/enclave/issues/332

Background
-  #318 has introduced initial OpenTelemetry setup
- #372 Event bus with telemetry support through

Goal:

- Extend our OpenTelemetry implementation to support observability backends. 
- As the project is in a development phase, the implementation should avoid any vendor lock-in.
- Minimal changes to existing code files to keep the code readability aspect intact.
- Possibly explore the self-hosted option of any tool before moving to the cloud version.
- possibly Implement a standalone OTEL collector. Currently, for Jaeger directly get the telemetry from the endpoint. The endpoint can be used to forward to an OpenTelemetry Collector(then to a backend like Jaeger or Signoz)

Possible options:

1. Direct integration with Signoz(via OLTP)

2. For sending local use promtail agent as a global service(all nodes)(https://grafana.com/docs/loki/latest/send-data/promtail/), then forward it to either Signoz(https://signoz.io/blog/using-opentelemetry-loki-receiver-to-collect-logs/#prerequisites) or Grafana-based receiver. requires using to existing `SimpleLogger` to write structured logs to files.

0xjei:
should we move this to later phases? @ryardley

PR #463, opened 2025-06-02
[DRAFT] signoz cloud otel collector support 

Subhasish-Behera:
Cloud support is complete
#432


Subhasish-Behera:
will add the version of running the entire setup locally without signoz cloud.

PR #478, opened 2025-06-05
add missing `hacknet` branch for releases action

### LINKED ISSUE #466: Ensure `packages/env` is published before folks can use `enclave init`
Currently as a work around I was using verdaccio locally to test `enclave init` but we should publish the package live.

0xjei:
this should ensure that #466 is satisfied without waiting for `main` merge (supports `hacknet` now)


0xjei:
after chatting with @ryardley - let's stick with `main` for now and iterate back whenever we have things ready.

PR #514, opened 2025-06-16
Update Docs

hmzakhalid:


0xjei:
@AllenAJ is there something needed here or we are okay to merge?

0xjei:
> Yeah lets merge this we should avoid long running branches if possible.

agree, I also reviewed the docs and lgtm!

PR #522, opened 2025-06-17
Expore the support image port

hmzakhalid:


crates/support-scripts/ctl/container
```diff
@@ -33,8 +33,8 @@ if docker ps -q -f name="$CONTAINER_NAME" | grep -q .; then
 else
     docker run -it --rm \
         --name "$CONTAINER_NAME" \
-        --network=host \
         --platform linux/amd64 \
+        -p 13151:13151 \
```
ryardley, line 37: --network=host exports the port automatically but is required so that processes running inside docker can see local processes outside of docker.

PR #589, opened 2025-07-23
Chore: lock compiler + deps and enable MSRV‑aware resolution

hmzakhalid:
* Pin Rust 1.86 via `rust-toolchain.toml`, `workspace.msrv`, and `rust-version` in every crate.  
* Switch resolver to v3 and add `--locked` to all build commands, freezing Cargo.lock.  
This guarantees reproducible builds and shields CI from surprise MSRV or API bumps in upstream crates.

## Summary by CodeRabbit

* **Chores**
  * Added the `--locked` flag to all Rust build and install commands across scripts, workflows, Dockerfiles, and documentation to ensure consistent dependency versions.
  * Updated Cargo workspace and package configurations to use resolver version "3" and set the minimum supported Rust version to "1.86.0".
  * Specified Rust toolchain version as "1.86.0" in relevant configuration files.
  * Revised documentation and build instructions to reflect these changes for improved build reproducibility.
  * Updated certain contract constants and file path references for improved accuracy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -399,7 +399,9 @@ jobs:
           toolchain: 1.86.0
           targets: wasm32-unknown-unknown
       - name: Install wasm-pack
-        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+        run: |
+          export VERSION=v0.13.1
+          curl -sSf https://raw.githubusercontent.com/alexcrichton/wasm-pack/master/docs/_installer/init.sh | sh
       - name: Install node dependencies
         run: pnpm install
       - name: Build the sdk


hmzakhalid:
You're right that the main issue with only pinning top-level dependencies is that subdependencies can still float. But:

* Semver does promise compatibility for patch and minor updates (unless it's a pre-1.0 crate, which is more lenient). So as long as crates follow semver, we should be safe.
* What actually broke in our case with `alloy` wasn't a semver violation, but an MSRV change and that’s not governed by semver.

We can now guard against MSRV related breakage using the resolver = "3" and the `msrv` field in `Cargo.toml`. That ensures incompatible versions won't be selected during resolution.

For CI, we'll still enforce `--locked` for build consistency, to catch any accidental lockfile drift.

And for local development, my plan is to pin all top-level versions using `=`, which gives us:

* Full control over version changes
* Zero ambiguity
* The ability to safely run `cargo build` without worrying about version upgrades

We’ll handle subdependency pinning on an as-needed basis using `[patch]` if it turns out to be a recurring problem  but semver + `Cargo.lock` should handle 95% of the stability we need.

ryardley:
> Semver does promise compatibility for patch and minor updates (unless it's a pre-1.0 crate, which is more lenient). So as long as crates follow semver, we should be safe.

Ok if you are cool with for now but part of me feels like we are placing a lot of trust in semver 😆 and I believe from a security perspective using exact versions will help with an audit - however one could also argue the exact versions are in the lock file.

ryardley:
> And for local development, my plan is to pin all top-level versions using =, which gives us:

Yes this is the way!

ryardley (approved):
🚀

PR #594, opened 2025-07-28
fix: wasm install

ryardley:
- wasm install fix
- bad pnpm lock file
- incorrect shebang for system compatability

## Summary by CodeRabbit

* **Chores**
  * Updated CI workflows to use a new, reusable local action for installing wasm-pack.
  * Replaced previous wasm-pack installation methods with a streamlined local installation step in CI jobs.

### DIFF
--- a/.github/actions/install-wasm-pack/action.yml
+++ b/.github/actions/install-wasm-pack/action.yml
@@ -1,0 +1,20 @@
+# NOTE: there has been a change and the rustwasm org is no more and the repository moved which
+# seems to have broken various things so this is an attempt to install directly from the binary download on github
+name: "Install wasm-pack"
+description: "Install wasm-pack from GitHub releases"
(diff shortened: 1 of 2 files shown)


ryardley:
🧙‍♂️ 📜 ✨ It came to pass that when the ancient forge-tool wasm-pack failed in its purpose, great joy arose among the Dwarfvelopers, for they were freed from their wearisome delving in the treacherous mines of npm and could instead find pleasant solace prospecting in the well-trodden, sun-lit hollows of Github, where treasures lay openly scattered for all to find. ⚒️ 🏔️ 💎

hmzakhalid (approved):
May we never have to touch this code again

ryardley:
> May we never have to touch this code again

So the villagers preyed before the days of the DRAGON!

Issue #597, opened 2025-07-29
Create a `Sensitive<T>` container for sensitive data.


ryardley:
We need a container to hold sensitive information in memory without being subject to memory dump and swapfile attacks this is especially important when sending sensitive data between actors.

Create a `Sensitive<T>` container that is designed to do the following:

- Be clonable and hold the data using an `Arc`
- Encrypt data on construction by passing in a `Cipher` instance.
- Ensure data is immutable
- Allow access to the data by unencrypting passing in a cipher instance.

PR #603, opened 2025-07-30
feat: add sensitive container for transferring data between actors

ryardley:
Closes: #597 

This makes the type system indicate sensitive data that can be decrypted with the cipher.

The reason for this is that we need to pass sensitive data such as secret shares to and from actors and this helps in the following ways:

- Type system prevents accidental logging of raw passwords
- Only encrypted data sits in actor mailboxes and memory
- Zeroizing auto-clears decrypted data when done
- Can't pass around plaintext since API forces encryption
- Failures only expose encrypted bytes, not passwords

### DIFF
--- a/crates/crypto/src/lib.rs
+++ b/crates/crypto/src/lib.rs
@@ -6,5 +6,6 @@
 
 mod cipher;
 mod password_manager;
+mod sensitive;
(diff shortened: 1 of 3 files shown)


0xjei (approved):
utACK!

PR #612, opened 2025-08-05
feat: event definition

ryardley:


crates/events/src/correlation_id.rs
```diff
@@ -9,10 +9,12 @@ use std::{
     sync::atomic::{AtomicUsize, Ordering},
 };
 
+use serde::{Deserialize, Serialize};
```
0xjei, line 12: are you expecting `serde` also on `trbfv` as well? there we were following the `protobuf` style which was a default of the lib.

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,149 @@
...
+    GenDecryptionShare(gen_decryption_share::Response),
+    ThresholdDecrypt(threshold_decrypt::Response),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum TrBFVError {
+    // Add errors here as required
+}
+
+pub mod gen_esi_sss {
```
0xjei, line 38: this will include the generation of smudging noise + secret shares from the derived poly, right?

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Response {
+        /// The smudging noise shares
+        esi_sss: Vec<SensitiveBytes>,
+    }
+}
+
+pub mod gen_pk_share_and_sk_sss {
```
0xjei, line 60: +1 on this

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,149 @@
...
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Response {
+        /// PublicKey share for this node
+        pk_share: Bytes,
+        /// SecretKey Shamir Shares for other parties
+        sk_sss: Vec<SensitiveBytes>,
+    }
+}
+
+pub mod gen_decryption_key {
```
0xjei, line 82: this is for the `sk` and `es` poly sums - are we missing the `pk` aggregation?

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,149 @@
...
+        es_poly_sum: SensitiveBytes,
+    }
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Response {
+        /// The decryption share for the given ciphertext
+        d_share_poly: Bytes,
+    }
+}
+
```
0xjei, line 128: are we missing `d_share_polys` aggregation before threshold decrypting?

0xjei (commented):
just left a few comments! thx @ryardley looks pretty well structured and self-explanatory

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,149 @@
...
+    use e3_crypto::SensitiveBytes;
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Request {
+        /// BFV Params
+        params: Bytes,
+        /// Max number of ciphertexts
+        max_num_ciphertexts: u64,
+    }
```
hmzakhalid, line 51: How are we passing the party information to the actor? It needs to know `num_parties` & `threshold` as well for `trbfv` and `ShareManager`.

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// This has been separated from the esi setup in order to be able to take advantage of parallelism
+    use crate::bytes::Bytes;
+    use e3_crypto::SensitiveBytes;
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Request {
+        /// BFV Params
+        params: Bytes,
+    }
```
hmzakhalid, line 71: Same here, we're missing the party info
I guess we could have a `PartyInfo` struct 

```
pub struct PartyInfo {
    pub party_id: u32,
    pub num_parties: u32,
    pub threshold: u32,
}
```

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Response {
+        /// PublicKey share for this node
+        pk_share: Bytes,
+        /// SecretKey Shamir Shares for other parties
+        sk_sss: Vec<SensitiveBytes>,
+    }
+}
```
hmzakhalid, line 80: do we need some more metadata here? can’t tell which bytes go to whom.

crates/events/src/enclave_event/compute_request/trbfv.rs
hmzakhalid: +1

hmzakhalid (commented):
Thank @ryardley, happy with the structure you're going with. left a few comments

crates/events/src/enclave_event/compute_request/trbfv.rs
hmzakhalid: +1

crates/events/src/enclave_event/compute_request/trbfv.rs
ryardley: That occurs in another actor as it is not a cryptographic operation just aggregation

crates/events/src/correlation_id.rs
ryardley: This is a good point - it is necessary as all actor messages need to be serializable. I will go over this and work out a good way to integrate protobuf

crates/events/src/enclave_event/compute_request/trbfv.rs
ryardley: This is separate from the pubkey+sk_sss because it can be done in parallel

crates/events/src/enclave_event/compute_request/trbfv.rs
ryardley: Pk aggregation is already done in the current implementation and from what I can tell need not change much

crates/events/src/correlation_id.rs
0xjei: yes, thanks. I can try to remove protobufs and use serde where needed on fhe.rs

crates/events/src/enclave_event/compute_request/trbfv.rs
0xjei: amazing, thanks for the clarification @ryardley

crates/events/src/enclave_event/compute_request/trbfv.rs
ryardley: We actually don't need the same in both from what I can tell.

ryardley: Smudging noise only needs max num ciphertexts but here we need threshold etc.

crates/net/src/document_publisher.rs
```diff
@@ -0,0 +1,16 @@
...
+// TODO: Create DocumentPublisher actor
+// 1. Accept EnclaveEvent::CiphernodeSelected and store selected e3_ids in a blume filter
+// 1. Accept EnclaveEvent::PublishDocumentRequested
+//    Take the payload and convert to NetCommand::PublishDocument
+// 1. Accept NetEvent::DocumentPublishedNotification from NetInterface
+//    Determine if we are keeping track of the given e3_id based on DocumentMeta
+//    and the e3_id blume filter if so then issue a NetCommand::FetchDocument
+// 1. Receive the document from NetEvent::FetchDocumentSucceeded and convert to
+//    EnclaveEvent::DocumentReceived
+// 1. Accept NetEvent::FetchDocumentFailed and attempt to retry
```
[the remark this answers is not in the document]
ryardley: This is not meant to be implementation just an outline of what to do.

ryardley:
I think this is good to merge as it gives us a scaffold for further developing the two parts of this

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,156 @@
...
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Request {
+        /// BFV Params
+        params: Bytes,
+        /// Ciphertext to decrypt
+        ciphertext: Bytes,
+        /// All decryption shares from a threshold quorum of nodes
+        d_share_polys: Vec<Bytes>,
```
YounesTal1, line 148: I am not sure we can here identify the party's id corresponding to each `d_share_poly`. This is needed to know how to use the elements of `d_share_polys` so as to decrypt

YounesTal1: Also i think the ciphertext is no longer needed at this point (it will not be used at this point to do the computation needed to decrypt, it was already used), but we need to know what are we decrypting so perhaps it is till necessary to take it as input

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,156 @@
...
+        /// All collected smudging noise shamir shares
+        esi_sss_collected: Vec<SensitiveBytes>,
+    }
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Response {
+        /// A single summed polynomial for this nodes secret key.
+        sk_poly_sum: SensitiveBytes,
+        /// A single summed polynomial for this partys smudging noise
+        es_poly_sum: SensitiveBytes,
```
YounesTal1, line 107: There might be situations where we need to generate more than one smudging noise for the same secret. This is the case where the output of the computation is the decryption of more than one ciphertext (e.g., a voting application where each ciphertext holds the votes of one candidate). I am just mentioning this here in case this is relevant for how the modules are written

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Request {
+        /// BFV Params
+        params: Bytes,
+        /// Max number of ciphertexts
+        max_num_ciphertexts: u64, // called num_summed in the example
+                                  // NOTE: we are using u64 for these values instead of usize
+                                  // so that the protocol has specific size
+                                  // not based on architecture
```
YounesTal1, line 53: Indeed the value of `max_num_ciphertexts` is relevant for the smudging noise, but i think we can replace it by the size from which it has to be drawn from, so as to not make this application specific. Otherwise, it should also be fine to keep it like this (but it's only convenient for CRISP), then the module does the necessary computation using `max_num_ciphertexts` to determine the size of the smudging noise.

YounesTal1: Also i think `num_parties` and `threshold` are needed as well

crates/events/src/enclave_event/compute_request/trbfv.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+    pub struct Request {
+        /// BFV Params
+        params: Bytes,
+        /// Num parties
+        num_parties: u64,
+        /// Threshold
+        threshold: u64,
+    }
```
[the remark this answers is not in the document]
ryardley: correct

PR #615, opened 2025-08-06
chore: remove redundant workspaces directive

ctrlc03:
Cleanup the root package.json from its redundant workspaces directive (https://github.com/gnosisguild/enclave/blob/main/pnpm-workspace.yaml takes precedence) 
 

## Summary by CodeRabbit

* **Chores**
  * Updated project configuration by removing the "workspaces" field. No impact on app features or user experience.


0xjei (approved):
utACK

PR #635, opened 2025-08-18
feat: move wasm-crypto to the sdk

### LINKED ISSUE #585: Move `CRISP/wasm-crypto` to the SDK
Move `encrypt_vote` from CRISP wasm-crypto to `verifiable_encrypt_u64` within the the typescript sdk

Take the code in https://github.com/gnosisguild/enclave/blob/main/examples/CRISP/apps/wasm-crypto/src/lib.rs

- Turn it into a simple function
- Place it in an appropriate rust sdk  https://github.com/gnosisguild/enclave/blob/main/crates/sdk/src/lib.rs potentially create a new crate for greco related functions.
- Have it exposed here converting types to JS types: https://github.com/gnosisguild/enclave/blob/main/crates/wasm/src/lib.rs
- Implement it in https://github.com/gnosisguild/enclave/blob/main/packages/enclave-sdk/src/wasm.ts

## Why?

1. Users building applications do not want to deal with the hassle of managing wasm nor the hassle of dealing with lower level verifiable compute primitives such as greco.

2. We should refactor to a simple function call because wasm bindings get complex when dealing with classes objects and shared memory and one of the easiest ways to deal with this is to only expose functions.

ctrlc03:
fix #585 

- Remove encryption WASM function from CRISP and move to own crate.
- Update enclave-sdk to use these functions 
- Let SDK be configurable based on protocol and params (with suggested params)
- Simplify the wasm bindings to a single function returning JsValue 
- Update the template and tests
- Update CRISP usage

Note that Vercel deployment will fail as it requires published updated packages as vercel does not rebuild the monorepo

## Summary by CodeRabbit

* **New Features**
  * BFV protocol support with configurable params; simple and verifiable encryption (with circuit inputs/proofs) via EnclaveSDK and new proof pipeline.

* **Breaking Changes**
  * Deprecated standalone encryptNumber; use EnclaveSDK methods or bfvEncryptNumber/bfvVerifiableEncryptNumber. Init/export names adjusted for multi-target builds.

* **Documentation**
  * Updated WASM README and examples to reflect new function names and signatures.

* **Tests**
  * Added SDK unit tests and comprehensive CRISP integration/e2e tests; Playwright Chromium project.

* **Chores**
  * CI/tools updated, packaging/export formats improved; removed legacy wasm-crypto.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -373,6 +373,9 @@ jobs:
           && sudo apt-get update -y \
           && sudo apt-get install -y solc
 
+      - name: Install wasm-pack
+        uses: ./.github/actions/install-wasm-pack
+
       - name: Install yq
         run: |
           sudo wget -qO- https://github.com/mikefarah/yq/releases/download/v4.45.1/yq_linux_amd64.tar.gz | sudo tar xz -C /usr/local/bin --transform='s|yq_linux_amd64|yq|' --no-same-owner
@@ -393,7 +396,8 @@ jobs:
 
       - name: Install dependencies
         working-directory: .
-        run: pnpm --version && pnpm install --frozen-lockfile
+        run: |
+          pnpm --version && pnpm install --frozen-lockfile && pnpm build:ts
 
       - name: Install Playwright browsers
         working-directory: ./examples/CRISP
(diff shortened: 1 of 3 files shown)


0xjei:
blocked by #657

0xjei (approved):
tACK - works on my end, feel free to merge once the CI green

0xjei (approved):

PR #669, opened 2025-09-02
feat: allow checkin without nargo

ryardley:
I noticed [nargo is now a dependency](https://github.com/gnosisguild/enclave/pull/648) in order to check in code because we use its linting. For this to work on my system I have to write a nix flake that builds nargo as it is [not available for nixos](https://search.nixos.org/packages?channel=25.05&query=nargo)  - this is going to take a bit of time for me so my initial reaction is to disable linting if nargo is not available

## Summary by CodeRabbit

* **Bug Fixes**
  * Circuit formatting check now gracefully skips when the required tool isn’t installed, preventing unnecessary failures.

* **Chores**
  * Minor cleanup of the linting script, including formatting improvements and clearer success messaging.

### DIFF
--- a/scripts/lint-circuits.sh
+++ b/scripts/lint-circuits.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
-
 set -euo pipefail
+
+# Check if nargo command exists
+if ! command -v nargo >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


ctrlc03 (approved):
LGTM

PR #676, opened 2025-09-04
fix(ci): ensure we don't have uncommited files

ctrlc03:
ensure we can release rust crates. failure: https://github.com/gnosisguild/enclave/actions/runs/17461577899/job/49587293099

## Summary by CodeRabbit

* Chores
  * Updated release automation to more reliably clean the workspace and exclude non-distributable example artifacts from release packaging. This reduces noise in release assets and prevents unintended files from being included. No functional changes to the product or APIs, and no action is required by users. Build and deployment processes remain the same, with improved consistency during release runs.


0xjei (approved):
easy

PR #698, opened 2025-09-09
Ticket Score Sortition

ozgurarmanc:


crates/sortition/src/sortition.rs
```diff
@@ -25,84 +28,243 @@ pub struct GetHasNode {
...
-            .into_iter()
-            // TODO: error handling
-            .map(|b| b.parse().unwrap())
-            .collect();
-
-        let Ok(committee) =
-            DistanceSortition::new(seed.into(), registered_nodes, size).get_committee()
-        else {
-            return Err(anyhow!("Could not get committee!"));
+    fn contains_distance(&self, seed: Seed, size: usize, address: &str) -> Result<bool> {
```
ryardley, line 110: Why are we differentiating here? would it simplify things to make the backends simply implement the trait?

```rust
impl SortitionList<String> for DistanceBackend {...}
impl SortitionList<String> for ScoreBackend {...}
impl SortitionList<String> for SortitionBackend {...}
```

```rust
enum SortitionBackend {
   Distance(DistanceBackend),
   Score(ScoreBackend)
}
```

```rust
impl SortitionList<String> for SortitionBackend {
   fn contains(&self, seed: Seed, size: usize, address: &str) -> Result<bool> {
     match self {
       Self::Distance(strategy) -> strategy.contains(seed, size, address),
       Self::Score(strategy) -> strategy.contains(seed, size, address),
     }
   }
   
   // etc...
}
```


An alternative to this is to use dynamic dispatch instead of an enum which is a little slower but means you have less matching code everywhere

```rust
struct SortitionBackend {
  strategy: Box<dyn SortitionList<String>>
}

impl SortitionList<String> for SortitionBackend {
   fn contains(&self, seed: Seed, size: usize, address: &str) -> Result<bool> {
     self.strategy.contains(seed, size, address)
   }
   
   // etc...
}

File: crates/sortition/src/index.rs [hunk 2]
```diff
@@ -6,14 +6,14 @@
 
 use rand::{rngs::StdRng, Rng, SeedableRng};
 
-pub struct IndexSortition {
+pub struct _IndexSortition {
```
ryardley, line 9: if we don't use this maybe we should delete it?

crates/sortition/src/sortition.rs
```diff
@@ -25,84 +28,243 @@ pub struct GetHasNode {
...
+
+        let committee =
+            DistanceSortition::new(seed.into(), registered_nodes, size).get_committee()?;
+
         Ok(committee
             .iter()
             .any(|(_, addr)| addr.to_string() == address))
     }
 
+    fn contains_score(&self, seed: Seed, size: usize, address: &str) -> Result<bool> {
```
ryardley, line 134: seems like this has the same api as `contains_distance`

crates/sortition/src/sortition.rs
```diff
@@ -25,84 +28,243 @@ pub struct GetHasNode {
...
+        };
+        Self { sortition, backend }
+    }
+
+    pub fn sort_type(&self) -> SortitionType {
+        self.sortition
+    }
+
+    /// Switch algorithm and migrate storage shape.
+    pub fn set_sort_type(&mut self, sort_type: SortitionType) {
```
ryardley, line 185: Seems like we need to think about this before implementing a switchover algorhythm. I think we should not support this at this point until we have a live system that actually needs updating on the fly.

crates/sortition/src/sortition.rs
```diff
@@ -25,84 +28,243 @@ pub struct GetHasNode {
...
+            sortition: SortitionType::Distance,
+            backend: SortitionBackend::default(),
+        }
+    }
+}
+
+impl SortitionModule {
+    /// Create a module with a specific algorithm.
+    pub fn new(sortition: SortitionType) -> Self {
+        let backend = match sortition {
```
ryardley, line 169: If we specify the backend type in the configuration would it be easier to just expect an enumerated backend in construction? 

thinking:

```rust
pub fn new(backend: SortitionBackend) -> Self;
```

```rust
#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct SortitionModule {
    backend: SortitionBackend,
}
```

ryardley (commented):
Main thing for me with this is questioning if we need to be able to swap strategies on the fly at this point.

crates/sortition/src/sortition.rs
ryardley: related to above

crates/sortition/src/sortition.rs
```diff
@@ -13,110 +15,328 @@ use e3_events::{
...
+/// for Distance, or a `Vec<RegisteredNode>` for Score), but they must be able to:
+/// - Check committee membership (`contains`)
+/// - Add and remove nodes
+/// - List all registered node addresses
 pub trait SortitionList<T> {
+    /// Return `true` if `address` appears in the size-`size` committee under `seed`.
+    ///
+    /// Implementations should return `Ok(false)` if the backend has no nodes
+    /// or if `size == 0`.
     fn contains(&self, seed: Seed, size: usize, address: T) -> Result<bool>;
```
ryardley, line 63: Not a suggestion for this PR but a thought: A short coming of this trait is that we may want to consider changing this to a predicate clsure in the future if we consider `SortitionList<RegisteredNode>`

```rust
if list.contains(seed, size, |n:&RegisteredNode| n.address == address)? {
  //..
}

// also 

let Some(party_id) = list.get_index(seed, size, |n:&RegisteredNode| n.address == address);
```

That might mean that we could add a `put` function to overwrite `RegisteredNode`s and then we could ditch the highly specific `set_ticket_count_addr()` and would be left wit a pretty simple API.

crates/sortition/src/sortition.rs
```diff
@@ -157,56 +385,69 @@ impl Handler<EnclaveEvent> for Sortition {
...
+                .or_insert_with(SortitionBackend::default_distance)
+                .add(addr);
             Ok(list_map)
         }) {
-            Err(err) => self.bus.err(EnclaveErrorType::Sortition, err),
-            _ => (),
-        };
+            self.bus.err(EnclaveErrorType::Sortition, err);
+        }
     }
```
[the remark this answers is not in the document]
ryardley: For now I would leave it the way it is but ditch `default_distance` and simply use `default()` which returns a `Distance` variant.

ryardley (commented):
I am considering we might want to change the trait in the future like I am suggesting but that does not have to happen here. Perhaps consider sorting out the default_distance -> default rename and lets just get this in. I can update my PR to accommodate the get_index stuff as I might want to change that a little too.

crates/sortition/src/sortition.rs
```diff
@@ -13,110 +15,320 @@ use e3_events::{
...
+                        address: addr,
+                        tickets: Vec::new(),
+                    });
+                }
+            }
+            Err(e) => {
+                tracing::warn!("Failed to parse address '{}': {}", address, e);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ryardley: Yeah we should not be doing this we should be adding RegisteredNode structs instead of Addresses

ryardley (approved):

PR #734, opened 2025-09-15
feat: optimization by concatenating coefficients

ozgurarmanc:
# Description
This PR will optimize Greco circuit by concatenating coefficients of the given polynomials.

Tests shows that this implementation lowers hashing payload costs nearly %66.

# PR Checklist\*
- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `nargo fmt` on default settings.

### DIFF
--- a/packages/circuits/crates/libs/greco/src/lib.nr
+++ b/packages/circuits/crates/libs/greco/src/lib.nr
@@ -4,7 +4,7 @@
 // without even the implied warranty of MERCHANTABILITY
 // or FITNESS FOR A PARTICULAR PURPOSE.
 
-use polynomial::{flatten, Polynomial};
+use polynomial::{flatten, packer, Polynomial};
 use safe::SafeSponge;
 
 /// Cryptographic parameters for Greco circuit.
@@ -329,7 +329,7 @@ impl<let N: u32, let L: u32> Greco<N, L> {
     /// # Returns
     /// Vector of challenge values [gamma_0, gamma_1, ..., gamma_{2L-1}]
     fn generate_challenge(self) -> Vec<Field> {
-        let inputs = self.payload::<(10 * N - 4) * L + 4 * N>();
+        let inputs = packer(self.payload::<((10 * N - 4) * L + 4 * N)>());
 
         // Domain separator for Greco circuit - "Greco" in hex
         let domain_separator = [
@@ -341,7 +341,7 @@ impl<let N: u32, let L: u32> Greco<N, L> {
         ];
 
         // IO Pattern: ABSORB(input_size), SQUEEZE(2*L)
-        let input_size = (10 * N - 4) * L + 4 * N;
+        let input_size = (((10 * N - 4) * L + 4 * N) + 2) / 3;
         let io_pattern = [0x80000000 | input_size, 0x00000000 | (2 * L)];
 
         let mut sponge = SafeSponge::start(io_pattern, domain_separator);
--- a/packages/circuits/crates/libs/polynomial/src/lib.nr
+++ b/packages/circuits/crates/libs/polynomial/src/lib.nr
@@ -132,6 +132,50 @@ pub fn flatten<let A: u32, let L: u32, let SIZE: u32>(
     (inputs, offset)
 }
 
+/// Packs an array of `Field` elements into a more compact representation,
+/// grouping three values into a single `Field` element. If the input length
+/// `A` is not divisible by 3, the final packed element is **zero-padded**.
+///
+/// # Type Parameters
+/// * `A` - Number of input coefficients.
+///
+/// # Parameters
+/// * `values` - The input array `[Field; A]` to be packed.
+///
+/// # Returns
+/// * `[Field; (A + 2) / 3]` - A packed array with length equal to
+///   `ceil(A / 3)`.
+///
+/// # Algorithm
+/// - Computes `S = ceil(A / 3)` (using `(A + 2) / 3` for integer division).
+/// - Iterates over each output position `j`:
+///   * Initializes an accumulator `acc = 0`.
+///   * For each of the 3 input slots belonging to this output position:
+///     * Reads `values[idx]` where `idx = 3 * j + i`, or uses `0` if `idx >= A`
+///       (to avoid out-of-bounds access).
+///     * Adds an offset of `2^64` to every value (ensuring positive packing).
+///     * Updates `acc = acc * 2^65 + adjusted_value`.
+/// - Stores `acc` into the output array at position `j`.
+///
+/// # Padding Behavior
+/// If `A % 3 != 0`, the last packed element will include one or two
+/// zero-padded values, ensuring that `values[idx]` is never accessed
+/// out-of-bounds.
+pub fn packer<let A: u32>(values: [Field; A]) -> [Field; (A + 2) / 3] {
+    let mut carrier = [0; (A + 2) / 3];
+    for j in 0..(A + 2) / 3 {
+        let base: Field = 36893488147419103232; // 2^65
+        let mut acc: Field = 0;
+        for i in 0..3 {
+            let idx: u32 = 3 * j + i;
+            let v: Field = if idx < A { values[idx] } else { 0 };
+            acc = acc * base + (v + 18446744073709551616); // + 2^64
+        }
+        carrier[j] = acc;
+    }
+    carrier
+}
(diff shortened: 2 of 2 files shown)


packages/circuits/crates/libs/greco/src/lib.nr
```diff
@@ -163,7 +163,7 @@ impl<let N: u32, let L: u32> Params<N, L> {
 /// * `k1` - Scaled message polynomial
 /// * `r1is`, `r2is` - Randomness polynomials for each i-th CRT basis.
 /// * `p1is`, `p2is` - Randomness polynomials for each i-th CRT basis.
-pub struct Greco<let N: u32, let L: u32> {
+pub struct Greco<let N: u32, let L: u32, let BIT_PK: u32, let BIT_U: u32, let BIT_E: u32, let BIT_K: u32, let BIT_R1: u32, let BIT_R2: u32, let BIT_P1: u32, let BIT_P2: u32> {
```
0xjei, line 166: can you add some comments on the `# Generic Parameters` section above for `BIT_PK`, `BIT_U` etc.

packages/circuits/crates/libs/greco/src/lib.nr
```diff
@@ -233,27 +233,27 @@ impl<let N: u32, let L: u32> Greco<N, L> {
...
         // Flatten public key and ciphertext polynomials first (public inputs)
-        let (inputs, offset) = flatten(inputs, self.pk0is, offset);
-        let (inputs, offset) = flatten(inputs, self.pk1is, offset);
-        let (inputs, offset) = flatten(inputs, self.ct0is, offset);
-        let (inputs, offset) = flatten(inputs, self.ct1is, offset);
+
+        inputs = flatten::<_, _, BIT_PK>(inputs, self.pk0is);
+        inputs = flatten::<_, _, BIT_PK>(inputs, self.pk1is);
+        inputs = flatten::<_, _, BIT_PK>(inputs, self.ct0is);
+        inputs = flatten::<_, _, BIT_PK>(inputs, self.ct1is);
```
0xjei, line 244: why do we have the same `BIT_PK` for `pk` and `ct`? if it's the same size, can we rename it to something meaningful for both?

packages/circuits/crates/libs/greco/src/lib.nr
```diff
@@ -233,27 +233,27 @@ impl<let N: u32, let L: u32> Greco<N, L> {
     ///
```
0xjei, line 233: is the `SIZE` comment still valuable and correct? I'm asking since we removed the `(10 * N - 4) * L + 4 * N` from the payload and we are using the length of the input

packages/circuits/crates/libs/polynomial/src/lib.nr
```diff
@@ -98,38 +98,87 @@ impl<let N: u32> Polynomial<N> {
...
-/// * `poly` - Array of polynomials to flatten
-/// * `offset` - Starting position in the inputs array
-///
-/// # Returns
-/// A tuple containing:
-/// * The updated inputs array with flattened coefficients
-/// * The new offset position after flattening
-pub fn flatten<let A: u32, let L: u32, let SIZE: u32>(
-    mut inputs: [Field; SIZE],
+/// Flatten `L` polynomials (each `Polynomial<A>`) into a linear Vec<Field> of packed carriers.
```
0xjei, line 101: can we improve a bit the docs here? I am having difficulties on understanding the functions

packages/circuits/crates/libs/polynomial/src/lib.nr
```diff
@@ -98,38 +98,87 @@ impl<let N: u32> Polynomial<N> {
...
+/// - Preserves deterministic order: iterate polynomials in `j`, then append all their carriers.
+pub fn flatten<let A: u32, let L: u32, let BIT: u32>(
+    mut inputs: Vec<Field>,
     poly: [Polynomial<A>; L],
-    mut offset: u32,
-) -> ([Field; SIZE], u32) {
+) -> Vec<Field> {
+    // Compute packing layout (must mirror `packer`).
+    let nibble_bits = ((BIT + 3) / 4) * 4; // ceil to multiple of 4
+    let limb_bits = nibble_bits + 4; // +1 nibble offset
```
0xjei, line 112: like, this might be a bit cryptic to read later on if need to make further changes

0xjei: like someone might ask why doing `() / 4) * 4`

packages/circuits/crates/libs/polynomial/src/lib.nr
```diff
@@ -98,38 +98,87 @@ impl<let N: u32> Polynomial<N> {
...
 
-    (inputs, offset)
+    inputs
+}
+
+/// Pack `values` into as few `Field` carriers as possible (hex-aligned).
+/// - `BIT`: bit width of each value (before offset). We round it up to a nibble.
+/// - Limb = `v + 2^nibble_bits` (lifts negatives); radix per limb = `2^(nibble_bits+4)`.
+/// - Packs `group = 254 / (nibble_bits+4)` values per carrier, big-endian; pads tail with `2^nibble_bits`.
+pub fn packer<let A: u32, let BIT: u32>(values: [Field; A]) -> Vec<Field> {
```
0xjei, line 137: same as above. I think for this kind of changes we should aim to have more "formal" documentation in the form of code comments or, if not enough, some kind of markdown - wdyt @auryn-macmillan

packages/circuits/crates/libs/polynomial/src/lib.nr
```diff
@@ -98,38 +98,87 @@ impl<let N: u32> Polynomial<N> {
...
+    assert(BIT >= 1);
+
+    // Align to hex digits (4 bits) and add +1 nibble headroom.
+    let nibble_bits = ((BIT + 3) / 4) * 4; // ceil_to_nibble(BIT)
+    let limb_bits = nibble_bits + 4; // limb width used for packing
+    assert(limb_bits <= 254);
+
+    // Max limbs per 254-bit field element.
+    let group = 254 / limb_bits;
+    assert(group >= 1);
```
0xjei, line 147: these lines are repeated in L110-L116. can we extract them in a meaningful and reusable function?

packages/circuits/crates/libs/safe/src/lib.nr
```diff
@@ -111,7 +111,7 @@ impl<let L: u32> SafeSponge<L> {
     ///
     /// # Arguments
     /// - `input`: Array of field elements to absorb (variable length, must match IO pattern)
-    pub fn absorb(&mut self, input: [Field]) {
+    pub fn absorb(&mut self, input: Vec<Field>) {
```
0xjei, line 114: generally speaking, do we wanna default to `Vec<Field>` instead of `[Field]` in most of the code? What are the tradeoffs there? I'm asking since this change in the poly lib in Noir might affect other circuits

0xjei (commented):
tACK. Left some comments. I think there are almost mandatory (imho) updates to code docs and some things to discuss but overall I'm okay if this gets merged as is. Therefore, I don't want to block the amazing work here and put in a comment.

note: I'm wondering if this packing can be moved in a separated Noir library (iff we are likely want to reuse it in other circuits).

packages/circuits/crates/libs/greco/src/lib.nr
ozgurarmanc: Instead, I added `BIT_CT` to avoid confusion there. I believe this will be a better solution to increase readability since it won't affect anything in the circuit. Wdyt?

packages/circuits/crates/libs/greco/src/lib.nr
ozgurarmanc: Deleted those comments in the latest commit. Thanks for spotting them!

packages/circuits/crates/libs/polynomial/src/lib.nr
ozgurarmanc: Tried to follow that formal approach with the latest commit!

0xjei (approved):
I am assuming this has ready to be merged. If anything arises from  https://github.com/gnosisguild/pvss/issues/17 for the `SIZE` I would prefer to tackle in another PR. thanks @ozgurarmanc for the amazing work here

PR #752, opened 2025-09-19
fix: fix reentrancy issue in enclave contracts

### LINKED ISSUE #751: Add reentrancy protection and CEI pattern to Enclave contracts
Several core functions (`request`, `publishInput`, `publishCiphertextOutput`, `publishPlaintextOutput`) currently perform external calls before updating contract state. This breaks the Checks-Effects-Interactions (CEI) pattern and introduces potential reentrancy risks, as flagged by static analysis.

A reentrancy guard on those functions should solve the issue, but as a best practice the code should still follow the Checks-Effects-Interactions (CEI) pattern.

cedoor:
Fixes #751 

## Summary by CodeRabbit

- New Features
  - None.

- Bug Fixes
  - Hardened critical user actions against reentrancy for requests and output publishing.
  - Fixed input indexing and hashing to ensure consistent event indices and stable hashes for consumers.
  - Ensured ciphertext and plaintext outputs are recorded before verification to improve state consistency during publish flows.

- Chores
  - Initialized and upgraded guard/security state to improve contract robustness.

### DIFF
--- a/packages/enclave-contracts/contracts/Enclave.sol
+++ b/packages/enclave-contracts/contracts/Enclave.sol
@@ -13,12 +13,15 @@ import {
     OwnableUpgradeable
 } from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import {
+    ReentrancyGuardUpgradeable
+} from "@openzeppelin/contracts-upgradeable/utils/ReentrancyGuardUpgradeable.sol";
+import {
     InternalLeanIMT,
     LeanIMTData,
     PoseidonT3
 } from "@zk-kit/lean-imt.sol/InternalLeanIMT.sol";
 
-contract Enclave is IEnclave, OwnableUpgradeable {
+contract Enclave is IEnclave, OwnableUpgradeable, ReentrancyGuardUpgradeable {
     using InternalLeanIMT for LeanIMTData;
 
     ////////////////////////////////////////////////////////////
@@ -114,6 +117,7 @@ contract Enclave is IEnclave, OwnableUpgradeable {
         bytes[] memory _e3ProgramsParams
     ) public initializer {
         __Ownable_init(msg.sender);
+        __ReentrancyGuard_init();
         setMaxDuration(_maxDuration);
         setCiphernodeRegistry(_ciphernodeRegistry);
         setE3ProgramsParams(_e3ProgramsParams);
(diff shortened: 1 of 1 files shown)


packages/enclave-contracts/contracts/Enclave.sol
```diff
@@ -296,7 +302,7 @@ contract Enclave is IEnclave, OwnableUpgradeable {
         uint256 e3Id,
         bytes calldata plaintextOutput,
         bytes calldata proof
-    ) external returns (bool success) {
+    ) external nonReentrant returns (bool success) {
```
ctrlc03, line 305: I would suggest not using the guard here and instead just following the checks-effect-interaction patter. That would achieve the same result but not incur any costs to the users.

i.e. move 
`e3s[e3Id].ciphertextOutput = ciphertextOutputHash;` before the external call to `e3Program.verify` and even if e3Program was a malicious contract and re-entered, there would be nothing they could do. Furthermore, e3 programs are whitelisted using `enableE3Program`, thus are not arbitrary external contracts. 

The same logic should be applied to `publishPlaintextOutput` -> move the state update `e3s[e3Id].plaintextOutput = plaintextOutput;` before the external call to the decryption verifier

packages/enclave-contracts/contracts/Enclave.sol
```diff
@@ -243,7 +247,7 @@ contract Enclave is IEnclave, OwnableUpgradeable {
     function publishInput(
         uint256 e3Id,
         bytes calldata data
-    ) external returns (bool success) {
+    ) external nonReentrant returns (bool success) {
```
ctrlc03, line 250: I'd probably refrain from using the guard here as well. What are the risks? let's imagine a request for an e3 is accepted (let's remember that all of these parameters are public at the time that a e3 request is made, even before a committee is selected) with a input validator contract that performs some reentrancy to the Enclave contract, what could they be doing? 

The contract re-enters and submits another input with different data, the data would be validated and inserted in the tree. That is no different than calling the publishInput function twice by two different addresses. 

Another option is that they re-enter with the same data, then `inputs[e3Id]._insert(inputHash);` will fail because you cannot insert two equal leaves into a LeanIMT. 

Anything I'm missing?

ctrlc03 (commented):
Left some comments

packages/enclave-contracts/contracts/Enclave.sol
cedoor: Yep, you’re right, I’ll remove the unnecessary parts. My reasoning was: “let’s add the guard” as an extra security layer. But if we apply the CEI pattern correctly, we can remove the guards and save some gas. The important part is making sure the CEI pattern continues to be followed consistently in the future.

Same for the other review.

PR #766, opened 2025-09-30
feat: ciphernode economic contracts

hmzakhalid:
## Summary by CodeRabbit

- New Features
  - USDC-based payments for E3 requests, with on-chain fee quotes.
  - Reward distribution to registered operators via a new Bonding Registry.
  - Operator lifecycle and eligibility managed through bonding, tickets, and licenses.
  - Slashing governance with proposals, verification, appeals, and bans.
  - New tokens: EnclaveToken (governance) and EnclaveTicketToken (non-transferable tickets).
  - Expanded registry APIs for committees, filters, and node management.
- Chores
  - Added deployment scripts and Ignition modules for new contracts.
  - New CLI tasks: admin add, mint tokens; updated package scripts.

## Summary by CodeRabbit

- **New Features**
  - ERC20 fee payments and fee-quoting; on-chain reward distribution; BondingRegistry, SlashingManager, EnclaveToken, EnclaveTicketToken; exit-queue and committee/filter APIs; slashing, banning, appeals, and reward distribution workflows; mock verifier and test stable token.

- **Tools**
  - New CLI tasks (admin-add, mint tokens), Ignition modules, deploy-and-save scripts, and updated deployment wiring to support token/registry flows.

- **Documentation**
  - Updated examples, env/config, and E3 configuration values.

- **Tests**
  - Expanded end-to-end and unit tests covering bonding, slashing, registry, and enclave flows.


packages/enclave-contracts/contracts/lib/ExitQueueLib.sol
```diff
@@ -0,0 +1,340 @@
...
+        address operator
+    ) internal view returns (uint256 ticketAmount, uint256 licenseAmount) {
+        PendingAmounts storage pending = state.pendingTotals[operator];
+        return (pending.ticketAmount, pending.licenseAmount);
+    }
+
+    function previewClaimableAmounts(
+        ExitQueueState storage state,
+        address operator
+    ) internal view returns (uint256 ticketAmount, uint256 licenseAmount) {
```
ctrlc03, line 141: not really familiar with writing libraries, assume it's ok to have view here because it's not modifying its own state?

packages/enclave-contracts/contracts/registry/NaiveRegistryFilter.sol
```diff
@@ -85,14 +89,13 @@ contract NaiveRegistryFilter is IRegistryFilter, OwnableUpgradeable {
 
     function publishCommittee(
         uint256 e3Id,
-        address[] calldata nodes,
-        bytes calldata publicKey
+        address[] memory nodes,
```
ctrlc03, line 92: why change these back to memory?

packages/enclave-contracts/contracts/token/EnclaveTicketToken.sol
```diff
@@ -0,0 +1,186 @@
...
+        IERC20 underlyingUSDC,
+        address registry_,
+        address initialOwner_
+    )
+        ERC20("Enclave Ticket Token", "ETK")
+        ERC20Permit("Enclave Ticket Token")
+        ERC20Wrapper(underlyingUSDC)
+        Ownable(initialOwner_)
+    {
+        require(registry_ != address(0), ZeroAddress());
```
ctrlc03, line 55: maybe shorten this by making setRegistry public and call it here?

packages/enclave-contracts/contracts/token/EnclaveToken.sol
```diff
@@ -0,0 +1,233 @@
...
+        uint256[] memory amounts,
+        string[] memory allocations
+    ) external onlyRole(MINTER_ROLE) {
+        if (
+            recipients.length != amounts.length ||
+            amounts.length != allocations.length
+        ) revert ArrayLengthMismatch();
+
+        uint256 totalAmount = 0;
+        for (uint256 i = 0; i < amounts.length; i++) {
```
ctrlc03, line 127: IIRC it's cheaper to cache the length to a variable first and pass it to the loop description rather than passing x.length directly

packages/enclave-contracts/contracts/token/EnclaveToken.sol
```diff
@@ -0,0 +1,233 @@
...
+        transferWhitelisted[account] = whitelisted;
+        emit TransferWhitelistUpdated(account, whitelisted);
+    }
+
+    /**
+     * @notice Toggle an account's whitelist status.
+     * @dev Only the owner may call this.
+     * @param account Address whose whitelist status should be toggled.
+     */
+    function toggleTransferWhitelist(address account) external onlyOwner {
```
ctrlc03, line 174: do we need setTransferWhitelist if we have this toggle function that can do both?

packages/enclave-contracts/tasks/ciphernode.ts
```diff
@@ -10,59 +10,496 @@ import { poseidon2 } from "poseidon-lite";
...
   })
   .setAction(async () => ({
-    default: async ({ ciphernodeAddress }, hre) => {
-      const { deployAndSaveCiphernodeRegistryOwnable } = await import(
-        "../scripts/deployAndSave/ciphernodeRegistryOwnable"
+    default: async ({ privateKey, licenseBondAmount, ticketAmount }, hre) => {
+      const connection = await hre.network.connect();
+      const { ethers } = connection;
+
+      if (!privateKey) {
```
ctrlc03, line 38: why passing the key as an arg when we can use the first signer generated by hardhat with our private key in either a .env or better eventually using the keystore https://hardhat.org/docs/learn-more/configuration-variables#how-to-use-configuration-variables-and-the-keystore-plugin ?

ctrlc03 (commented):
Great work :) left some comment, will give it another pass tomorrow as there's some parts of the contracts I wanted to spend some more time on

packages/enclave-contracts/contracts/Enclave.sol
```diff
@@ -324,9 +350,39 @@ contract Enclave is IEnclave, OwnableUpgradeable {
...
         emit PlaintextOutputPublished(e3Id, plaintextOutput);
     }
 
+    ////////////////////////////////////////////////////////////
+    //                                                        //
+    //                   Internal Functions                   //
+    //                                                        //
+    ////////////////////////////////////////////////////////////
+
+    function _distributeRewards(uint256 e3Id) internal {
```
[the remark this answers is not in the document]
auryn-macmillan: Worth noting that LLMs are particularly good at drafting out natspec comments

packages/enclave-contracts/contracts/Enclave.sol
```diff
@@ -324,9 +350,39 @@ contract Enclave is IEnclave, OwnableUpgradeable {
...
+        IRegistryFilter.Committee memory committee = ciphernodeRegistry
+            .getCommittee(e3Id);
+        uint256[] memory amounts = new uint256[](committee.nodes.length);
+
+        // We might need to pay different amounts to different nodes.
```
auryn-macmillan, line 369: Make sure to leave "TODO" in the comment so others can easily find things like this that we know will need to be changed in future.

packages/enclave-contracts/contracts/Enclave.sol
```diff
@@ -437,6 +518,12 @@ contract Enclave is IEnclave, OwnableUpgradeable {
         return InternalLeanIMT._root(inputs[e3Id]);
     }
 
+    function getE3Quote(
+        E3RequestParams calldata
+    ) public pure returns (uint256 fee) {
+        fee = 1 * 10 ** 6;
```
auryn-macmillan, line 524: Perhaps needs a "TODO"

packages/enclave-contracts/contracts/lib/ExitQueueLib.sol
auryn-macmillan: The function doesn't actually modify any state, just reads state, so it can be marked as view.

ctrlc03: yea right it modifies `operatorQueue ` which is  part of that `state` storage variable - wasn't really sure how it'd work with that actual being a storage update on some other contract (not own ofc as libraries do not hold any state)

auryn-macmillan: iiuc, `operatorQueue` is a storage pointer. The storage slot that it points to is not modified. Similar for `tranche`.

packages/enclave-contracts/contracts/registry/NaiveRegistryFilter.sol
hmzakhalid: ahh, it got copied over from the previous branch, it wasnt up to date with `main`

hmzakhalid:
Thanks @auryn-macmillan & @ctrlc03 for the quick review. I've pushed the changes!

packages/enclave-contracts/contracts/registry/NaiveRegistryFilter.sol
```diff
@@ -118,7 +121,9 @@ contract NaiveRegistryFilter is IRegistryFilter, OwnableUpgradeable {
 
     function getCommittee(
         uint256 e3Id
-    ) external view returns (Committee memory) {
-        return committees[e3Id];
+    ) external view returns (IRegistryFilter.Committee memory) {
+        IRegistryFilter.Committee memory committee = committees[e3Id];
+        require(committee.publicKey != bytes32(0), CommitteeNotPublished());
+        return committee;
     }
```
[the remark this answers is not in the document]
hmzakhalid: Custom reverts work in new versions of solidity

packages/enclave-contracts/contracts/Enclave.sol
```diff
@@ -197,6 +220,9 @@ contract Enclave is IEnclave, OwnableUpgradeable {
         e3.decryptionVerifier = decryptionVerifier;
 
         e3s[e3Id] = e3;
+        e3Payments[e3Id] = e3Fee;
+
+        usdcToken.safeTransferFrom(msg.sender, address(this), e3Fee);
```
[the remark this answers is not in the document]
hmzakhalid: I would remove it, since the plan is to initially accept stable token and later on use the ENCL token, we shouldn't keep it payable if it's not going to accept any native tokens.

packages/enclave-contracts/contracts/Enclave.sol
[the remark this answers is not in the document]
hmzakhalid: Ill do a review of all contracts and add the natspec if we're missing them. we should probably use @inheritdoc in the implementations and add the natspec across all interfaces.

packages/enclave-contracts/contracts/lib/ExitQueueLib.sol
```diff
@@ -0,0 +1,340 @@
...
+            if (lastTranche.unlockTimestamp == unlockTimestamp) {
+                if (ticketAmount != 0) lastTranche.ticketAmount += ticketAmount;
+                if (licenseAmount != 0)
+                    lastTranche.licenseAmount += licenseAmount;
+                merged = true;
+            }
+        }
+
+        if (!merged) {
+            operatorQueue.push();
```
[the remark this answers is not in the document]
hmzakhalid: yeah it just appends a zero-initialized ExitTranche, we can replace it with this
```
if (!merged) {
    ExitTranche storage t = operatorQueue.push();
    t.unlockTimestamp = unlockTimestamp;
    t.ticketAmount = ticketAmount;
    t.licenseAmount = licenseAmount;
}
```
a bit shorter and doesn't indexes with len

packages/enclave-contracts/contracts/token/EnclaveToken.sol
```diff
@@ -0,0 +1,217 @@
...
+    error TransferNotAllowed();
+
+    /// @dev Maximum supply of the token (18 decimals).
+    uint256 public constant TOTAL_SUPPLY = 1_200_000_000e18;
+
+    /// @dev Role allowing accounts to mint new tokens.
+    bytes32 public constant MINTER_ROLE = keccak256("MINTER_ROLE");
+
+    /// @dev Tracks the amount of tokens minted so far.
+    uint256 public totalMinted;
```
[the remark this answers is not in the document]
hmzakhalid: yeah we can drop it if we dont plan to support burns (which i think we will? @auryn-macmillan), but id assume then `MAX_SUPPLY` is a cap on current supply.

but if If `MAX_SUPPLY` is a cap on ever minted (burns shouldn’t let us remint) then we need to keep `totalMinted` since `totalSupply` would let us remint new tokens after a burn.

packages/enclave-contracts/contracts/slashing/SlashingManager.sol
```diff
@@ -0,0 +1,342 @@
...
+        SlashPolicy storage policy = slashPolicies[reason];
+        require(policy.enabled, SlashReasonDisabled());
+
+        proposalId = totalProposals;
+        bool proofVerified = false;
+
+        if (policy.requiresProof) {
+            require(proof.length != 0, ProofRequired());
+            proofVerified = ISlashVerifier(policy.proofVerifier).verify(
+                proposalId,
```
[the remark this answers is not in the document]
hmzakhalid: not 100% sure what the exact verifier logic will look like (or where spent nullifiers would be tracked), but I think of proposalId here the same way I’d think of nullifiers. it’s basically replay protection. we bind the proof to a specific proposal so it cant be reused.

that said, we’ll also need to include things like the job id (e3Id) and the chainId for cross chain slashing, so a proof can’t be replayed across different proposals or chains.

packages/enclave-contracts/contracts/Enclave.sol
```diff
@@ -126,12 +260,12 @@ contract Enclave is IEnclave, OwnableUpgradeable {
...
+    /// @inheritdoc IEnclave
     function request(
         E3RequestParams calldata requestParams
-    ) external payable returns (uint256 e3Id, E3 memory e3) {
-        // TODO: allow for other payment methods or only native tokens?
-        // TODO: should payment checks be somewhere else? Perhaps in the E3 Program or ciphernode registry?
-        require(msg.value > 0, PaymentRequired(msg.value));
+    ) external returns (uint256 e3Id, E3 memory e3) {
+        uint256 e3Fee = getE3Quote(requestParams);
+        require(e3Fee > 0, PaymentRequired(e3Fee));
```
ctrlc03, line 268: is this to prevent some sort of param combination that might cause getE3Quote to to return zero? perhaps it could be moved inside getE3Quote directly?

packages/enclave-contracts/contracts/registry/BondingRegistry.sol
```diff
@@ -0,0 +1,757 @@
...
+    }
+
+    /// @inheritdoc IBondingRegistry
+    function isActive(address operator) external view returns (bool) {
+        return operators[operator].active;
+    }
+
+    /// @inheritdoc IBondingRegistry
+    function hasExitInProgress(address operator) external view returns (bool) {
+        Operator storage op = operators[operator];
```
ctrlc03, line 283: no need for storage here

packages/enclave-contracts/contracts/registry/BondingRegistry.sol
```diff
@@ -0,0 +1,757 @@
...
+    /// @dev Restricts function access to only the slashing manager
+    modifier onlySlashingManager() {
+        if (msg.sender != slashingManager) revert Unauthorized();
+        _;
+    }
+
+    /// @dev Reverts if operator has an exit in progress that hasn't unlocked yet
+    /// @param operator Address of the operator to check
+    modifier noExitInProgress(address operator) {
+        Operator storage op = operators[operator];
```
ctrlc03, line 130: can be memory

packages/enclave-contracts/contracts/registry/BondingRegistry.sol
```diff
@@ -0,0 +1,757 @@
...
+    /// @inheritdoc IBondingRegistry
+    function setExitDelay(uint64 newExitDelay) public onlyOwner {
+        uint256 oldValue = uint256(exitDelay);
+        exitDelay = newExitDelay;
+
+        emit ConfigurationUpdated("exitDelay", oldValue, uint256(newExitDelay));
+    }
+
+    /// @inheritdoc IBondingRegistry
+    function setSlashedFundsTreasury(
```
ctrlc03, line 668: this set function and the ones below do not have events like the others. Won't we need them?

packages/enclave-contracts/contracts/registry/BondingRegistry.sol
```diff
@@ -0,0 +1,757 @@
...
+        );
+        require(!operators[msg.sender].registered, AlreadyRegistered());
+        require(
+            operators[msg.sender].licenseBond >= licenseRequiredBond,
+            NotLicensed()
+        );
+
+        operators[msg.sender].registered = true;
+
+        if (address(registry) != address(0)) {
```
ctrlc03, line 311: perhaps we should add the check inside setRegistry instead? so it can never be zero and we avoid the check here for every call by ciphernodes? or are there cases where we wouldn't have a ciphernode registry set?

packages/enclave-contracts/contracts/lib/ExitQueueLib.sol
```diff
@@ -0,0 +1,485 @@
...
+     * @param state The exit queue state storage
+     * @param operator The operator to query
+     * @return ticketAmount Total pending tickets in the exit queue
+     * @return licenseAmount Total pending licenses in the exit queue
+     */
+    function getPendingAmounts(
+        ExitQueueState storage state,
+        address operator
+    ) internal view returns (uint256 ticketAmount, uint256 licenseAmount) {
+        PendingAmounts storage pending = state.pendingTotals[operator];
```
ctrlc03, line 221: not sure we need storage here?

packages/enclave-contracts/contracts/lib/ExitQueueLib.sol
```diff
@@ -0,0 +1,485 @@
...
+     * @param state The exit queue state storage
+     * @param operator The operator to query
+     * @return ticketAmount Total claimable tickets at current timestamp
+     * @return licenseAmount Total claimable licenses at current timestamp
+     */
+    function previewClaimableAmounts(
+        ExitQueueState storage state,
+        address operator
+    ) internal view returns (uint256 ticketAmount, uint256 licenseAmount) {
+        ExitTranche[] storage operatorQueue = state.operatorQueues[operator];
```
ctrlc03, line 237: also here we don't need storage for trance and operatorQueue right?

packages/enclave-contracts/contracts/lib/ExitQueueLib.sol
```diff
@@ -0,0 +1,485 @@
...
+     * @return ticketsClaimed Actual amount of tickets claimed
+     * @return licensesClaimed Actual amount of licenses claimed
+     */
+    function claimAssets(
+        ExitQueueState storage state,
+        address operator,
+        uint256 maxTicketAmount,
+        uint256 maxLicenseAmount
+    ) internal returns (uint256 ticketsClaimed, uint256 licensesClaimed) {
+        if (maxTicketAmount > 0) {
```
ctrlc03, line 269: I think we can remove these if (ticket/license amount > 0) because inside _takeAssetsFromQueue we already check if it's 0

packages/enclave-contracts/contracts/slashing/SlashingManager.sol
```diff
@@ -0,0 +1,370 @@
...
+            policy.licensePenalty,
+            executableAt,
+            msg.sender
+        );
+
+        totalProposals = proposalId + 1;
+    }
+
+    /// @inheritdoc ISlashingManager
+    function executeSlash(uint256 proposalId) external onlySlasher {
```
ctrlc03, line 253: perhaps anyone should be allowed to call this if they wish to?

packages/enclave-contracts/contracts/slashing/SlashingManager.sol
```diff
@@ -0,0 +1,370 @@
...
+        emit AppealFiled(proposalId, p.operator, p.reason, evidence);
+    }
+
+    /// @inheritdoc ISlashingManager
+    function resolveAppeal(
+        uint256 proposalId,
+        bool appealUpheld,
+        string calldata resolution
+    ) external onlyGovernance {
+        require(proposalId < totalProposals, InvalidProposal());
```
ctrlc03, line 333: do we need this if we have `require(p.appealed, InvalidProposal());` ? if they provide an invalid index, we'd have an empty proposal where appealed would be set to the default boolean value false

packages/enclave-contracts/contracts/slashing/SlashingManager.sol
```diff
@@ -0,0 +1,370 @@
...
+            resolution
+        );
+    }
+
+    // ======================
+    // Ban Management
+    // ======================
+
+    /// @inheritdoc ISlashingManager
+    function banNode(address node, bytes32 reason) external onlyGovernance {
```
ctrlc03, line 356: what are the scenarios where we'd ban a node without slashing them first? maybe keeping a list of permabanned addresses (sanctioned ones for instance) that we'd never want to allow to be ciphernodex?

crates/events/src/enclave_event/committee_published.rs
```diff
@@ -0,0 +1,30 @@
...
+use actix::Message;
+use serde::{Deserialize, Serialize};
+use std::fmt::{self, Display};
+
+#[derive(Message, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[rtype(result = "()")]
+pub struct CommitteePublished {
+    pub e3_id: E3id,
+    pub nodes: Vec<String>,
+    pub public_key: Vec<u8>,
```
ryardley: Maybe this should this be `ArcBytes`? That might mean we end up being more memory efficient and can just use the Debug output in the Display formatter like this:

```rust
impl Display for CommitteePublished {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "{:?}", self)
    }
}
```

Without worrying about a flood of data

crates/keyshare/src/threshold_keyshare.rs
```diff
@@ -767,13 +771,64 @@ impl Handler<EnclaveEvent> for ThresholdKeyshare {
...
         match self.handle_ciphernode_selected(msg, ctx.address()) {
             Err(e) => error!("{e}"),
             Ok(_) => (),
         }
     }
 }
 
+impl Handler<CommitteeFinalized> for ThresholdKeyshare {
+    type Result = ();
+    fn handle(&mut self, msg: CommitteeFinalized, _: &mut Self::Context) -> Self::Result {
```
ryardley: private note for me look into what is happening here not sure i understand

crates/sortition/src/ciphernode_selector.rs
```diff
@@ -87,28 +100,118 @@ impl Handler<E3Requested> for CiphernodeSelector {
...
+            let meta_repo = repositories.meta(&e3_id);
+            let Some(e3_meta) = meta_repo.read().await.ok().flatten() else {
+                info!(
+                    node = address,
+                    "No stored E3 metadata for {:?}, skipping", e3_id
+                );
+                return;
+            };
+
+            let Some(party_id) = msg.committee.iter().position(|addr| addr == &address) else {
```
ryardley: Maybe we should have a Specific Committee struct for this?
```rust
let Some(party_id) = msg.committee.get_party_id(&address);
```

crates/sortition/src/node_state.rs
```diff
@@ -0,0 +1,383 @@
...
+    }
+}
+
+/// State for all nodes across all chains
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+pub struct NodeStateStore {
+    /// Map of (chain_id, node_address) to node state
+    pub nodes: HashMap<(u64, String), NodeState>,
+    /// Current ticket price per chain
+    pub ticket_prices: HashMap<u64, U256>,
```
ryardley: Stupid question: do we need to be concerned about what token this is represented in?

crates/aggregator/src/publickey_aggregator.rs
```diff
@@ -228,7 +213,8 @@ impl Handler<NotifyNetwork> for PublicKeyAggregator {
     fn handle(&mut self, msg: NotifyNetwork, _: &mut Self::Context) -> Self::Result {
         Box::pin(
             self.sortition
-                .send(GetNodes {
```
ryardley: I think this event is no longer used so it should be deleted

crates/sortition/src/sortition.rs
```diff
@@ -488,30 +758,48 @@ impl Handler<CiphernodeRemoved> for Sortition {
...
+        let node_state_manager = self.node_state_manager.clone();
+        let bus = self.bus.clone();
+
+        // Get the sortition backends synchronously
+        let backends_snapshot = self.list.get();
+
+        Box::pin(async move {
+            // Query NodeStateManager for fresh state
+            let node_state_snapshot = if let Some(manager) = node_state_manager {
+                manager.send(GetNodeState).await.ok().flatten()
```
ryardley: This seems really heavy handed to me. We just want an index and to do that we appear to get the entire NodeState global HashMap from the NodeStateManager that contains the state of every single node on every single chain and pass it to sortition. Actors are meant to only share data by message passing and we want to limit the size of the messsages we pass. To my eyes we should consider ditching `GetNodeState` in favour of a redesign.

crates/aggregator/Cargo.toml
```diff
@@ -14,6 +14,7 @@ bincode = { workspace = true }
 e3-config = { workspace = true }
 e3-data = { workspace = true }
 e3-events = { workspace = true }
+e3-evm = { workspace = true }
```
ryardley: I am not convinced we should share anything from evm with aggregator as aggregator is kind of a core module and evm is shell.

We should have a core version of the FinalizeCommittee event in events we use.

crates/aggregator/src/committee_finalizer.rs
```diff
@@ -0,0 +1,156 @@
...
+use std::collections::HashMap;
+use std::time::Duration;
+use tracing::{error, info};
+
+/// CommitteeFinalizer is an actor that listens to CommitteeRequested events and calls
+/// finalizeCommittee on the registry after the submission deadline has passed.
+pub struct CommitteeFinalizer {
+    #[allow(dead_code)]
+    bus: Addr<EventBus<EnclaveEvent>>,
+    registry_writer: Recipient<FinalizeCommittee>,
```
ryardley: Whilst using Recipient allows for polymorphic testing we are still passing in a direct reference. Instead of passing a direct reference like this I would suggest we consider dispatching a `ComitteeFinalizeRequested` event on the bus so that core and shell know nothing about each other and things are totally separated. We then don't need to do anything fancy on construction and can just pass the bus in

crates/keyshare/src/threshold_keyshare.rs
```diff
@@ -767,13 +771,63 @@ impl Handler<EnclaveEvent> for ThresholdKeyshare {
 impl Handler<CiphernodeSelected> for ThresholdKeyshare {
     type Result = ();
     fn handle(&mut self, msg: CiphernodeSelected, ctx: &mut Self::Context) -> Self::Result {
+        // Store selection for potential committee finalization
+        self.pending_selections
+            .insert(msg.e3_id.clone(), msg.clone());
+
+        // If CommitteeFinalized arrives later, it will verify committee membership
```
ryardley: `handle_ciphernode_selected` will start the key generation process which is very expensive - are we sure we want to do this before the committee has been finalized and we know we are part of the committee? Maybe this makes sense but it seems like it might be premature optimization.

crates/request/src/context.rs
```diff
@@ -38,6 +39,8 @@ pub struct E3Context {
     pub dependencies: HetrogenousMap,
     /// A Repository for storing this context's data snapshot
     pub repository: Repository<E3ContextSnapshot>,
+    /// Root data store for accessing global repositories
+    root_store: DataStore,
```
ryardley: why do we have a DataStore and a Repository here? Should we not be just adding to the E3Context?

crates/tests/tests/integration.rs
```diff
@@ -173,10 +204,40 @@ async fn test_trbfv_actor() -> Result<()> {
 
     bus.do_send(event);
 
-    // NOTE: We are using node 0 as the aggregator but it is not selected in this seed which is why
-    // there is no CiphernodeSelected event
+    // For score sortition, we need to wait for nodes to process E3Requested and run sortition
+    // Since TicketGenerated is a local-only event (not shared across network), we can't collect it
+    // we need to manually construct the committee that sortition would select
+    println!("Waiting for nodes to process E3Requested...");
+    tokio::time::sleep(Duration::from_millis(200)).await;
```
ryardley: It might be good to test for the events using `take_history_with_timeout()`? That should avoid us needing to add sleeps

crates/tests/tests/integration.rs
```diff
@@ -138,8 +142,35 @@ async fn test_trbfv_actor() -> Result<()> {
...
+        parameter: "ticketPrice".to_string(),
+        old_value: U256::ZERO,
+        new_value: U256::from(10_000_000u64),
+        chain_id,
+    }))
+    .await?;
+
     for node in nodes.iter() {
-        adder.add(&node.address()).await?;
+        let addr = node.address();
```
ryardley: should we consider using the `setup_score_sortition()` function in the other test?

crates/request/src/context.rs
hmzakhalid: Ahh, I wanted to address this. I was having issues tracking down the scope where E3Meta was stored at in the router since I couldn't access it for some reason, so to save time I added a global store (root_store) to test the functionality. 

I finally tracked down what was causing the mismatch between the scopes, since the scope `"//router//router//router//context//e3_id//meta//e3_id"` didnt make sense.

In `E3Router::build()`, the code calls `self.store.repositories().router()` but `self.store` is already a scoped to `"//router"`. so calling `.router()` again appends another `"//router"` creating `"//router//router"`
but this wasnt the weird part since i noticed the scope changed after restart. when recovering from snapshot in `E3Router::from_snapshot()`, the code called `repositories.router()` again on an already scoped store, adding yet another `"//router"` layer with each restart

I've removed the `root_store` and now use scopes in CiphernodeSelector 
```
  let repositories = self
            .data_store
            .scope(StoreKeys::router())
            .scope(StoreKeys::context(&e3_id))
            .repositories();
          
          
let meta_repo = repositories.meta(&e3_id);
```

Should I be using a `Repository<E3Meta>` instead? I definitely not using these correctly.

ryardley: Yes certainly

ryardley (approved):
**L**ets **G**et **T**his **M**erged 🚀

PR #850, opened 2025-10-14
refactor: remove error collector

ryardley:
### What is this change?

- Remove the `ErrorCollector` test helper because we are not using it very much and it can be done with the `HistoryCollector`
- Rename `GetHistory` & `TakeHistory` to `GetEvents` and `TakeEvents`

### Why?

The history collector got some [advanced functionality](https://github.com/gnosisguild/enclave/blob/ry/601-tidy-up-error-collector/crates/events/src/eventbus.rs#L346-L389) in `TakeEvents` and this was not reflected in the ErrorCollector instead of duplicating the code it seems better to just use a HistoryCollector and only listen for Error events. 

## Summary by CodeRabbit

* **Refactor**
  * Unified event history and error logs under a single Events interface for simpler, consistent access.
  * Streamlined naming across the platform (History → Events) without changing runtime behavior.
  * Consolidated error tracking into the standard event history for a single source of truth.

* **Tests**
  * Updated integration tests to use the new Events interface and naming, maintaining existing assertions and behavior.

### DIFF
--- a/examples/CRISP/sdk/package.json
+++ b/examples/CRISP/sdk/package.json
@@ -33,11 +33,12 @@
     "@types/node": "22.7.5",
     "chai": "^6.2.0",
     "prettier": "^3.2.5",
-    "vitest": "^1.6.1",
-    "typescript": "^5.0.0"
+    "typescript": "^5.0.0",
+    "vitest": "^1.6.1"
   },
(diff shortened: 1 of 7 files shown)


cedoor (approved):
utACK 👍🏽

PR #859, opened 2025-10-16
chore: calculate yes and no indexes in ts

ctrlc03:
Calculate the plaintext indexes given a degree and the total voting power. This is needed to ensure that the encoded plaintext sent to the greco circuit has the correct binary values set in the correct indices. That will prevent overflowing and malicious inputs. 

The total supply is used as this calculates the worst case scenario where one voter has the full voting power which would be the maximum vote possible we would need to fit. (in reality it's always less, but we can't assume a more correct value imo)

## Summary by CodeRabbit

* **New Features**
  * Added token total supply snapshot querying at specific blockchain blocks
  * Added voting index calculation utility for plaintext constraint validation

* **Tests**
  * Added comprehensive test coverage for voting index calculations

### DIFF
--- a/examples/CRISP/sdk/src/token.ts
+++ b/examples/CRISP/sdk/src/token.ts
@@ -73,7 +73,7 @@ export const getBalanceAt = async (voterAddress: string, tokenAddress: string, s
 }
 
 /**
- * Get the total supply of a ERC20 Token at a specific block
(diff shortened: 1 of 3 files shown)


cedoor:
@coderabbitai review

cedoor (approved):
utACK 👍🏽

PR #862, opened 2025-10-17
fix: pnpm-lock.yaml inconsistency

ryardley:
`pnpm-lock.yaml` appears to be inconsistent.

What I run here:

`pnpm install`
`pnpm clean`
`cd packages/enclave-contracts`
`pnpm hardhat compile`

## Summary by CodeRabbit

* **Chores**
  * Minor internal formatting adjustment with no user-facing impact or functional changes.


ryardley:
> Not sure this solves solves anything? I've run into this too. Seems related to the Git submodules.

Yeah I know what you mean - I just figure it does not hurt at all and can rule out version.

PR #894, opened 2025-10-24
fix: deploy risc0verifier with hardhat

ctrlc03:
## Summary by CodeRabbit

* **New Features**
  * Added a new mock contract for RiscZeroGroth16Verifier.

* **Refactor**
  * Updated deployment process to use ethers.js instead of Forge-based scripts.
  * Updated OpenZeppelin package dependency path resolution to use `node_modules` instead of library symlinks.


cedoor (approved):
utACK

PR #902, opened 2025-10-24
fix: wait until start window before activating an e3

ctrlc03:
Ensure we wait until the activation period starts before calling e3.activate

## Summary by CodeRabbit

* **Bug Fixes**
  * Contract activation now respects scheduled start times and will wait until the scheduled moment before activating.
  * Added logging to indicate whether activation was delayed and for how long, or if activation occurred immediately, improving synchronization transparency.


cedoor (approved):
utACK 👍🏽

PR #922, opened 2025-10-28
chore: update size check workflow

cedoor:
## Summary by CodeRabbit

* **Chores**
  * Improved PR size check to count changed lines (added + removed) across files and exclude lock files for more accurate sizing.
  * Clarified output messaging to indicate lock files are excluded.
  * Simplified the size-comparison logic to use the updated line count directly while preserving pass/fail behavior.


ctrlc03 (approved):
utACK

PR #936, opened 2025-10-29
fix: add some stability improvements to crisp_e2e test

ryardley:
I had a few issues running crisp on my local system this also helped fix CI which was also having issues with running in headless mode. This PR helps with fixing these issues.

1. Displaying a huge key starved my system of resources as it was doing proof processing so I used derivative to not print the proof and keys to the output (probably a "my system" issue but still helpful in general as we are trying to use this to avoid printing large keys in logging) 
2. Hot module reloading was causing the test to stall as there were multiple http://localhost:3000 requests which occasionally caused an error that failed the test so I disabled it in the crisp test by creating a client `pnpm dev-static` script that does not do hot module reloading.
3. Made a definitive reset and test script for a clean checkout to aid local testing:
    `./scripts/run-crisp-test.sh --ui` and `./scripts/run-crisp-test.sh`
4. Fixed an issue with `xvf-run` where I finally worked out that not providing a screen size was causing the test to stall both locally on CI.

## Summary by CodeRabbit

* **New Features**
  * Added dev-static script for static development mode without hot module reloading.
  * Added run-crisp-test script for executing comprehensive cleanup and E2E testing workflow.

* **Improvements**
  * Enhanced development script startup logging and synchronization with health checks.
  * Improved E2E test configuration with better display settings and reliability measures.

* **Chores**
  * Added workspace dependency for development tooling.

### DIFF
--- a/scripts/run-crisp-test.sh
+++ b/scripts/run-crisp-test.sh
@@ -1,3 +1,9 @@
 #!/usr/bin/env bash
 
+echo "This helper script will clean your repository and run end-to-end tests for CRISP."
+echo "WARNING: This will reset your current workspace. Ensure all changes are committed before proceeding."
+echo "Press any key to continue or Ctrl+C to cancel..."
+
+read
+
 rm -rf * && git reset --hard HEAD && git submodule update --init --recursive && pnpm install && cd examples/CRISP && pnpm test:e2e "$@"


ryardley:
Ok this is good to go

PR #965, opened 2025-11-05
fix: make submit ticket more gas efficient

hmzakhalid:
We were running out of gas during submitTicket (e.g., on the 4th ticket) because maintaining a sorted/top-N via storage shifts or O(N) scans is too expensive at scale.

Replaced sorted insert with an unsorted bounded set in `_insertTopN`. 
- When not full, we append once and store the `scoreOf` 
- When full, we scan to find current worst and replace in place if the new score is smaller.
- Keeps top N smallest scores.


ryardley:
I think this was merged before the crisp_e2e test passed.

PR #972, opened 2025-11-06
fix: change bound types in Greco to Field

ozgurarmanc:
## Summary by CodeRabbit

* **Refactor**
  * Range/bound parameters migrated from integer types to Field; error bound split into separate e0/e1 paths. Component now stores an additional polynomial collection and enforces distinct e0/e1 flattening and checks.
* **Tests**
  * Updated tests to use Field-typed bounds and positive lower-bound semantics; single-bound range check removed in favor of two-bound checks.
* **Examples**
  * Example circuits updated to supply the new polynomial collection to constructors.

### DIFF
--- a/examples/CRISP/circuits/src/ciphertext_addition.nr
+++ b/examples/CRISP/circuits/src/ciphertext_addition.nr
@@ -127,8 +127,8 @@ impl<let N: u32, let L: u32, let BIT_ZERO_CT: u32, let BIT_PREV_CT: u32, let BIT
         for i in 0..L {
             // For M=2 (adding two ciphertexts), each coefficient of the quotient polynomial.
             // must be in {-1, 0, 1}, so the bound will always be 1 for all CRT moduli.
-            self.r0is[i].range_check_1bound::<1>(1);
-            self.r1is[i].range_check_1bound::<1>(1);
+            self.r0is[i].range_check_2bounds::<1>(1, 1);
+            self.r1is[i].range_check_2bounds::<1>(1, 1);
         }
     }
 


circuits/crates/libs/polynomial/src/lib.nr
```diff
@@ -117,11 +117,11 @@ impl<let N: u32> Polynomial<N> {
     /// # Arguments
     /// * `upper_bound` - The symmetric bound for coefficient range checking
     ///                   Coefficients must satisfy: -upper_bound <= c <= upper_bound
-    pub fn range_check_1bound<let BIT: u32>(self, upper_bound: u64) {
-        let shifted_bound = (upper_bound as Field) + (upper_bound as Field);
+    pub fn range_check_1bound<let BIT: u32>(self, upper_bound: Field) {
```
cedoor: `range_check_1bound` could just call `range_check_2bounds` with two `upper_bound` parameters at this point no?

0xjei: right!

0xjei:
```rust
self.range_check_2bounds::<BIT>(upper_bound, upper_bound);    
```

you mean this @cedoor ?

so we can do like

```rust
self.u.range_check_2bounds::<BIT_U>(bound_params.u_bound, bound_params.u_bound);

// instead of
self.u.range_check_1bound::<BIT_U>(bound_params.u_bound);
```

0xjei: done in [46f29e1](https://github.com/gnosisguild/enclave/pull/972/commits/46f29e15175024422bfe804590c9488f374fd49e)

circuits/crates/libs/greco/src/lib.nr
```diff
@@ -182,14 +189,15 @@ pub struct Greco<let N: u32, let L: u32, let BIT_PK: u32, let BIT_CT: u32, let B
     u: Polynomial<N>,
     e0: Polynomial<N>,
     e1: Polynomial<N>,
+    e1is: [Polynomial<N>; L],
```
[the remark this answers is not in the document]
0xjei: we already have range check for `e1` we need to constrain `e1 mod qi == e1is[i]` but we are going to do it in another pr, great catch btw!

circuits/crates/libs/polynomial/src/lib.nr
cedoor: Yess

cedoor (approved):
utACK 👍🏽

PR #994, opened 2025-11-11
chore: add static version of sdk in client

cedoor:
## Summary by CodeRabbit

* **Refactor**
  * Simplified the build pipeline by removing the intermediate SDK build step and consolidating TypeScript compilation and bundling into a single flow.

* **Chores**
  * Pinned the SDK dependency to a specific release (0.2.0-test) instead of using a workspace reference.

### DIFF
--- a/examples/CRISP/client/vite.config.ts
+++ b/examples/CRISP/client/vite.config.ts
@@ -23,15 +23,7 @@ export default defineConfig({
   },
   optimizeDeps: {
     esbuildOptions: { target: 'esnext' },
-    exclude: [
-      '@rollup/browser',
-      '@crisp-e3/zk-inputs',
-      '@crisp-e3/sdk',
-      '@noir-lang/noirc_abi',
-      '@noir-lang/acvm_js',
-      '@noir-lang/noir_js',
-      '@aztec/bb.js',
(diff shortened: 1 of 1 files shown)


examples/CRISP/client/package.json
```diff
@@ -12,15 +12,14 @@
...
-    "build": "pnpm build:sdk && tsc && vite build",
+    "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "preview": "vite preview",
     "predeploy": "pnpm run build",
     "deploy": "gh-pages -d dist"
   },
   "dependencies": {
-    "@crisp-e3/sdk": "workspace:*",
+    "@crisp-e3/sdk": "0.2.0-test",
```
ryardley: Is doing this antithetical to having a monorepo? Does this not make testing much harder?

cedoor: I kept the same configuration but I see your point. Locally the client should use the local package as dependency, and ideally it should use the static version when deployed. We could open another issue for this maybe?

ctrlc03 (approved):
utACK

PR #1012, opened 2025-11-13
feat: add historical events ordering on ciphernode startup

hmzakhalid:
This PR introduces an Event Ordering Layer. 

Problem:
Currently, each EvmEventReader fetches historical events independently and sends them to the EventBus immediately so there's no coordination between readers from different contracts
Which means `CiphernodeRegistry` events (`CiphernodeAdded`) might arrive before `BondingRegistry` events (`OperatorActivationChanged`). so state can be inconsistent if events arrive out of order
Which caused nodes to be inactive during sortition.

## The fix

Added a `HistoricalEventCoordinator` that:
- Buffers all historical events from all readers
- Waits for all readers to finish fetching
- Sorts events by block number
- Publishes them to the EventBus in the correct order

Once historical sync is done, live events bypass the coordinator completely and go straight to the EventBus.
```
CiphernodeBuilder
  ├─ Counts readers (4 per chain: Enclave, BondingRegistry, CiphernodeRegistry, EnclaveSolReader)
  └─ Creates single HistoricalEventCoordinator with reader_count

All readers start concurrently
  ├─ BondingRegistry fetches logs → buffers events → signals complete
  ├─ CiphernodeRegistry fetches logs → buffers events → signals complete  
  ├─ EnclaveSolReader fetches logs → buffers events → signals complete
  └─ EnclaveSol fetches logs → buffers events → signals complete

Last reader to finish triggers coordinator:
  ├─ Sort all buffered events by block number
  ├─ Publish sorted events to EventBus
  └─ Log: "replay complete, published N ordered events"

Now all readers are in live mode:
  └─ New events published immediately 
```

## Summary by CodeRabbit

* **New Features**
  * Coordinated historical-event replay that buffers and orders past events before delivering live streams.

* **Bug Fixes**
  * More reliable operator activation handling.
  * Added warning when ticket price is zero.

* **Chores**
  * Updated deployed contract records to reflect refreshed addresses and proxy deployment details.

* **Tests**
  * Expanded integration tests covering coordinated historical-first processing, multi-reader scenarios, and resume-after-shutdown behavior.

### DIFF
--- a/crates/ciphernode-builder/src/ciphernode_builder.rs
+++ b/crates/ciphernode-builder/src/ciphernode_builder.rs
@@ -304,7 +304,6 @@ impl CiphernodeBuilder {
         let mut provider_cache = ProviderCaches::new();
         let cipher = &self.cipher;
 
-        // Setup coordinator for event ordering
         let coordinator = HistoricalEventCoordinator::setup(local_bus.clone());
         let processor = coordinator.clone().recipient();
 
--- a/crates/evm/tests/integration.rs
+++ b/crates/evm/tests/integration.rs
@@ -19,7 +19,7 @@ use e3_entrypoint::helpers::datastore::get_in_mem_store;
 use e3_events::{
     new_event_bus_with_history, EnclaveEvent, GetEvents, HistoryCollector, Shutdown, TestEvent,
 };
-use e3_evm::{helpers::EthProvider, EvmEventReader, HistoricalEventCoordinator, Start};
+use e3_evm::{helpers::EthProvider, CoordinatorStart, EvmEventReader, HistoricalEventCoordinator};
 use std::time::Duration;
 use tokio::time::sleep;
 
@@ -95,7 +95,7 @@ async fn evm_reader() -> Result<()> {
     )
     .await?;
 
-    coordinator.do_send(Start);
+    coordinator.do_send(CoordinatorStart);
 
     contract
         .setValue("hello".to_string())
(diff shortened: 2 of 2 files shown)


crates/evm/src/historical_event_coordinator.rs
```diff
@@ -0,0 +1,79 @@
...
+    block: u64,
+    event: EnclaveEvent,
+}
+
+/// Coordinates historical replay across all EvmEventReaders.
+/// Buffers historical events, then sorts + publishes once all readers finish.
+pub struct HistoricalEventCoordinator {
+    pending_readers: AtomicUsize,
+    events: Mutex<Vec<BufferedEvent>>,
+    bus: Addr<EventBus<EnclaveEvent>>,
```
ryardley: Mutex Atomics... looks like this aught to be an Actor.

crates/ciphernode-builder/src/ciphernode_builder.rs
```diff
@@ -303,6 +303,37 @@ impl CiphernodeBuilder {
...
+                reader_count += 1;
+            }
+            if self.contract_components.enclave_reader {
+                reader_count += 1;
+            }
+            if self.contract_components.bonding_registry {
+                reader_count += 1;
+            }
+            if self.contract_components.ciphernode_registry {
+                reader_count += 1;
```
ryardley: Seems redundant to do this when we do it below.

crates/evm/tests/integration.rs
```diff
@@ -73,6 +73,7 @@ async fn evm_reader() -> Result<()> {
         &bus,
         &repository,
         rpc_url.clone(), // Pass RPC URL
+        None,
```
ryardley: It would be good to test this here

ryardley (approved):
Approving anyway to get this in but we should consider this as technical debt

PR #1015, opened 2025-11-13
chore: update vercel install command

### LINKED ISSUE #999: Add setup script for CRISP client deployment
At the moment, deploying the CRISP client requires a static version of the JS SDK, which undermines the main benefits of having CRISP in the monorepo (synchronized development and consistent dependency linking). We need to install toolchains (Rust, Nargo, wasm-pack) and build dependent packages, and ultimately automate this process with a script/CLI.

cedoor:
Closes #999 

Instead of adding a custom deployment script for Vercel, this PR introduces the pnpm `link-workspace-packages` option. This allows the workspace to link the local package even when a static version is defined. On Vercel, this option is simply disabled, so the published npm package is used instead.

It also adds `wasm-pack` as an npm dependency, which is far easier to maintain than relying on external binaries. A few script improvements were included as well to streamline the workflow.

## Summary by CodeRabbit

* **Chores**
  * Updated build and packaging scripts across examples to streamline circuit compilation and wasm build steps.
  * Added tooling and dev-dependencies to support wasm workflows.
  * Added project-specific install/configuration file to standardize installs.
  * Improved publish pipeline to update client dependency and enhance logging/error handling.
* **Style**
  * Minor formatting and stylistic cleanups in publish tooling.

### DIFF
--- a/examples/CRISP/client/vercel.json
+++ b/examples/CRISP/client/vercel.json
@@ -1,3 +1,3 @@
 {
-  "installCommand": "pnpm install --no-link-workspace-packages"
+  "installCommand": "pnpm install --no-link-workspace-packages --no-lockfile"
 }


examples/CRISP/scripts/publish.ts
cedoor: lots of lines bcs it was not formatted

PR #1025, opened 2025-11-18
refactor: bind encrypted vote to zk proof [skip-line-limit]

### LINKED ISSUE #983: Add conversion function from GRECO/Noir ciphertext to BFV format
We need a function to convert the return value from the CRISP proof, which contains the ciphertext to be stored in the slot, from the GRECO/Noir format to the BFV format (including re-centering and related adjustments).

This means the new vote will be an array of bytes32 of length $D \cdot 2 \cdot L$.

The function should be implemented in Rust and then used in the program before the computation.

cedoor:
The PR binds the circuit’s ciphertext output directly to the proof. Before this change, the ciphertext and the proof were disconnected, which made the proof effectively useless. Now the on-chain contract receives the exact ciphertext used during zk verification, encodes it, and emits it in the Ethereum event. The event type remains bytes, but the internal format is now the encoded (with `abi.encode`) greco ciphertext, and needs proper conversion off-chain.

On the CRISP program side, two new functions were added:
• `abi_decode_greco_ciphertext`: performs ABI decoding (bytes → GRECO ciphertext),
• `greco_to_bfv_ciphertext`: converts the ciphertext from GRECO format to BFV format.

These functions introduce additional cycles during the RISC0 verification, so checks were minimized and the implementations optimized.

This PR is the first step toward a transition where only the commitments of the circuit’s public inputs will be published on-chain, rather than the full polynomials. This means that instead of passing the ciphertext and the public-key row, only their hashes will be passed.

Closes #983

## Summary by CodeRabbit

* **New Features**
  * Greco→BFV conversion utilities and a Solidity-proof encoding function added.

* **Changes**
  * Vote flow unified to a single encoded_proof value across client, hooks, context, server, and models.
  * validateInput no longer returns the decoded input payload.
  * Verifier public-input size and verification key updated.

* **Documentation**
  * Input Validation guidance updated to revert on invalid input.

* **Chores**
  * Package versions bumped and new crypto/math helper dependencies added.

* **Tests**
  * New test deployment utilities and rewritten tests to use them.

### DIFF
--- a/examples/CRISP/packages/crisp-contracts/contracts/CRISPVerifier.sol
+++ b/examples/CRISP/packages/crisp-contracts/contracts/CRISPVerifier.sol
@@ -10,143 +10,143 @@ uint256 constant LOG_N = 18;
 uint256 constant NUMBER_OF_PUBLIC_INPUTS = 2066;
 uint256 constant VK_HASH = 0x2116432a8071b45d4f2e0cec74a985a1ea1de1008309715fdb59de0707087fc2;
 library HonkVerificationKey {
-    function loadVerificationKey() internal pure returns (Honk.VerificationKey memory) {
-        Honk.VerificationKey memory vk = Honk.VerificationKey({
-            circuitSize: uint256(262144),
-            logCircuitSize: uint256(18),
-            publicInputsSize: uint256(2066),
-            ql: Honk.G1Point({ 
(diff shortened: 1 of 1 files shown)


cedoor:
@hmzakhalid it would be nice to check the number of cycles in Risc0 after this PR.

PR #1026, opened 2025-11-18
chore: update Noir to latest version in CRISP [skip-line-limit]

### LINKED ISSUE #1013: Upgrade Noir noir_js and Aztec bb.js versions
We need to upgrade Noir noir_js and Aztec bb.js to the latest stable/nightly versions to stay aligned with the security fixes and protocol updates in the Aztec stack.

```
@noir-lang/noir_js → 1.0.0-beta.14
@aztec/bb.js → 3.0.0-nightly.20251030-2
```

cedoor:
Closes #1013 

Skipping the line limit as the new verifier is quite heavy.

## Summary by CodeRabbit

* **New Features**
  * Added protocol validation for encryption operations to ensure configuration correctness.

* **Bug Fixes**
  * Improved contract deployment architecture with proper library linking for verifier contracts.

* **Chores**
  * Updated toolchain versions (Nargo to beta.15, bb.js to 3.0.0-nightly.20251104).
  * Updated documentation and configuration examples with latest version requirements.
  * Refined optimizer settings for deployment efficiency.

### DIFF
--- a/examples/CRISP/client/vite.config.ts
+++ b/examples/CRISP/client/vite.config.ts
@@ -22,7 +22,7 @@ export default defineConfig({
   },
   optimizeDeps: {
     esbuildOptions: { target: 'esnext' },
-    exclude: ['@rollup/browser', '@noir-lang/noirc_abi', '@noir-lang/acvm_js', '@noir-lang/noir_js', '@aztec/bb.js'],
+    exclude: ['@rollup/browser', '@noir-lang/noirc_abi', '@noir-lang/acvm_js', '@noir-lang/noir_js'],
   },
   resolve: {
     alias: {


cedoor:
Should we get rid of the functions to generate/verify proof in the Enclave SDK? That is currently generating GRECO proofs without application logic (no checks on plaintext), which is not the way apps will build their circuits. So I'm not sure when that functionality can actually be useful for devs. Am I missing anything?

Also, the circuit which was used to generate the .json artifact for the sdk is no longer in our repo, as we built an actual CRISP circuit.

Something we could do is to provide functions to wrap noir/bb without passing any specific inputs and without using any specific circuit artifact, which will be passed by devs based on their circuits.

@ryardley @ctrlc03

packages/enclave-sdk/src/greco.ts
```diff
@@ -4,8 +4,8 @@
 // without even the implied warranty of MERCHANTABILITY
 // or FITNESS FOR A PARTICULAR PURPOSE.
 
-import { UltraHonkBackend, type ProofData } from '@aztec/bb.js';
-import { type CompiledCircuit, Noir } from '@noir-lang/noir_js';
+import { UltraHonkBackend, type ProofData } from "@aztec/bb.js";
+import { type CompiledCircuit, Noir } from "@noir-lang/noir_js";
```
ryardley: at some point we aught to fail for bad linting like this. (not this PR)

cedoor:
> yea this makes sense. Perhaps simply encryption proving with greco could be more of a guideline for users rather than something they'd use directly

https://github.com/gnosisguild/enclave/issues/1032

PR #1027, opened 2025-11-18
feat: store merkle tree on program [skip-line-limit]

### LINKED ISSUE #947: Update tally verification to use votes in slots rather than enclave level inputs merkle tree

### LINKED ISSUE #981: Extract merkle tree for inputs from Enclave to the program level
Currently, we store all inputs sent by users via Enclave.publishInput inside a merkle tree. The plan is to make this application specific logic (program) rather than baked into the enclave protocol. 

For instance, there might be applications where not all inputs should be considered, such as CRISP where only the latest vote of a user is to be considered.

### LINKED ISSUE #1037: Fix setRoundData flag scope (should be per-e3, not program-level)
`setRoundData` in the CRISP program is currently storing a flag at program level to prevent data from being stored more than once. This is incorrect.

The flag should be tracked per e3, otherwise one e3 call blocks all others and leads to incorrect behavior.

ctrlc03:
Closes #981
Closes #947
Fixes #1037 

## Summary by CodeRabbit

* **New Features**
  * Per-E3 scoped round and vote state; programs now emit InputPublished per E3. SDK can expose a PublicClient instance.
* **Bug Fixes**
  * Indexer/repository persist ciphertext inputs and listeners updated to consume program-emitted InputPublished.
* **Breaking Changes**
  * Public APIs updated to include an e3Id parameter. Enclave no longer exposes input root/length getters or emits InputPublished; related event/types removed from SDK/react exports and tests adjusted.

### DIFF
--- a/examples/CRISP/packages/crisp-contracts/contracts/Mocks/MockCRISPProgram.sol
+++ b/examples/CRISP/packages/crisp-contracts/contracts/Mocks/MockCRISPProgram.sol
@@ -29,6 +29,9 @@ contract MockCRISPProgram is IE3Program, Ownable {
     // Constants
     bytes32 public constant ENCRYPTION_SCHEME_ID = keccak256("fhe.rs:BFV");
 
+    // The depth of the input merkle tree
+    uint8 public constant TREE_DEPTH = 20;
+
     // State variables
     IEnclave public enclave;
     IRiscZeroVerifier public verifier;
@@ -205,7 +208,7 @@ contract MockCRISPProgram is IE3Program, Ownable {
         returns (bool)
     {
         require(paramsHashes[e3Id] != bytes32(0), E3DoesNotExist());
-        bytes32 inputRoot = bytes32(votes[e3Id]._root());
+        bytes32 inputRoot = bytes32(votes[e3Id]._root(TREE_DEPTH));
         bytes memory journal = new bytes(396); // (32 + 1) * 4 * 3
 
         encodeLengthPrefixAndHash(journal, 0, ciphertextOutputHash);


examples/CRISP/packages/crisp-contracts/contracts/CRISPProgram.sol
```diff
@@ -26,6 +29,9 @@ contract CRISPProgram is IE3Program, Ownable {
     // Constants
     bytes32 public constant ENCRYPTION_SCHEME_ID = keccak256("fhe.rs:BFV");
 
+    // The depth of the input merkle tree
+    uint8 public constant treeDepth = 20;
```
cedoor: Should we use upper case for this?

cedoor (changes requested):

packages/enclave-contracts/package.json
```diff
@@ -174,6 +173,7 @@
   "dependencies": {
     "@openzeppelin/contracts-upgradeable": "^5.0.2",
     "@zk-kit/lean-imt.sol": "2.0.1",
+    "@zk-kit/lean-imt": "2.2.4",
```
cedoor: Do we need this dependency?

crates/indexer/tests/integration.rs
```diff
@@ -61,56 +61,6 @@ async fn test_indexer() -> Result<()> {
         .watch()
         .await?;
 
-    // InputPublished
```
ryardley: Might be good to replace this with a different event. This just happened to use this one as an example to test the indexer.

.github/workflows/ci.yml
```diff
@@ -689,6 +689,7 @@ jobs:
       - name: Test Template
         run: |
           cd templates/default
+          pnpm compile
```
ryardley: seems odd to me this had not happened before - perhaps a smell that we aught to get turbo repo working

crates/compute-provider/src/merkle_tree.rs
```diff
@@ -6,24 +6,57 @@
...
 use ark_ff::{BigInt, BigInteger};
-use lean_imt::LeanIMT;
 use light_poseidon::{Poseidon, PoseidonHasher};
 use num_bigint::BigUint;
 use num_traits::Num;
 use sha3::{Digest, Keccak256};
 use std::str::FromStr;
+use zk_kit_imt::imt::IMT;
 
 pub struct MerkleTree {
```
ryardley: Should this be called `MerkleTreeBuilder`?

crates/indexer/tests/integration.rs
cedoor: Makes sense!

cedoor (approved):

0xjei (approved):

Issue #1020, opened 2025-11-14
sync to blockchain time


ryardley:
As the protocol becomes more complex we are having timing issues within out end to end tests creep in.

we need to fastforward in sync with blockchain time.

We should do this by:

Making all "waits" dependent on blocks received from whatever chain we are listening to.

This means listening to each block and processing the block extracting the timestamp and considering that the timestamp to react to within the protocol.

Eg

```ts
test('full e2e with blockchain time travel', async ({ page }) => {
  // 1. Mock browser time
  await page.clock.install({ time: Date.now() });
  
  await page.goto('http://localhost:3000');
  
  // 2. Advance blockchain time (via test RPC)
  await page.evaluate(async () => {
    await window.ethereum.request({
      method: 'evm_increaseTime',
      params: [3600]
    });
    await window.ethereum.request({ method: 'evm_mine' });
  });
  
  // 3. Advance browser time to match
  await page.clock.fastForward(3600000); // Should probably not be required
  
  // Now both client and blockchain are synchronized in the future
  await page.click('#check-status');
  await expect(page.locator('#result')).toContainText('ready');
});
```

We should ensure all components synchronize with block advancement to make end to end tests run more quickly.

ryardley:
#1052 is ready to merge. Going to put this down and focus on event sourcing for now and come back to this once we have better stability in the node.

PR #1042, opened 2025-11-21
feat: add timestamp driven callbacks to indexer

ryardley:
References #1020 

- Marks sections to refactor
- Add new `ThresholdQueue` for executing threshold callbacks. We can use this later with events in ciphernodes too.
- Add specialized version for Indexer called `CallbackQueue`
- Setup block listener functionality in `EventListener`
- Ensured that on failure listening retries
- Created `dispatch_after_tijmestamp` method on `EnclaveIndexer`
- Colocated public methods
- Hid non public exports
- Extracted delayed functionliaty to `handle_e3_input_deadline_expiration` function
- Needed to run TS tools with pnpm
- Contract documentation was inaccurate compared to how we are using the expires field so I updated the docs.


ryardley:
putting this on hold until #1043 is merged

Issue #1050, opened 2025-11-25
Add sync for ciphernodes


ryardley:
- ~~lamport clock~~ hybrid logical clock on the event bus for all EnclaveEvents
- all eventbus events are stored as they are published (probably not in sled but in a commitlog) along with an index of the latest processed event 
- ~~latest processed event is an event that has had all the snapshots that have been caused as a consequence of it's propagation stored. This should probably just be an event broadcast like `EventProcessedSuccessfully(EventId=1243)`~~ might not be necessary
- ~~after a shutdown we only sync libp2p events from open e3 rounds. Maybe?~~ Possibly unnecessary.
- shutdown can happpen at anytime or due to panic
- every broadcast or write has a corresponding success or failure event that is published on the eventbus - will help withsync `EventProcessedSuccessfully(EventId=1243)`
- every event that is a libp2p event is identified and at the end of syncing there is a request for all events from ~~lamport~~ hlc time

ryardley:
@hmzakhalid do you have any other ideas here?

ryardley:
Gonna get going on this tomorrow.

ryardley:
```mermaid
sequenceDiagram
    participant HLC as HLC Clock
    participant EVM as EVM Module
    participant Blockchain as Blockchain
    participant Bus as Event Bus
    participant Internal as Internal Actors
    participant Net as Net Module
    participant P2P as P2P Network

    %% EVM Events - DETERMINISTIC HLC
    rect rgb(240, 240, 255)
        Note right of Blockchain: Source A: EVM Events (DETERMINISTIC)
        Blockchain->>EVM: Blockchain Event<br/>(block_num, tx_idx, log_idx)
        EVM->>EVM: Derive deterministic HLC<br/>from chain data
        EVM->>HLC: hlc.receive(deterministic_hlc)
        Note right of HLC: Update local clock to stay<br/>ahead of blockchain time
        EVM->>Bus: EnclaveEvent (with deterministic HLC)
    end

    %% P2P Events - REMOTE
    rect rgb(240, 255, 240)
        Note right of P2P: Source C: P2P Events (REMOTE)
        P2P->>Net: GossipData (contains remote HLC)
        Net->>HLC: hlc.receive(remote_hlc)
        HLC-->>Net: Updated HLC timestamp
        Net->>Bus: EnclaveEvent (with merged HLC)
    end

    %% Internal Events - LOCAL
    rect rgb(255, 240, 240)
[...]

PR #1153, opened 2026-01-08
feat: sync mode preparation [skip-line-limit]

ryardley:
part of #1050 

### This PR

We are part of the way through getting sync together. Much of this is preparatory. The next PR will be Sync implementation. 

#### Event writing system and snapshot:
- [x] Setup `EventContext` to attach context to `EnclaveEvent`. **Why?** so that we can have provenance between events for state reconstruction. Also we need to have the aggregate_id on the snapshots. 
- [x] Setup `TypedEvent<T>`. **Why?** so that we can disambiguate destructured events from the ctx in order to pass the context through to dispatchers.
- [x] Simplify `Persistable` by removing unnecessary DataStore and Repository flow. **Why?** persistable was far more complex than it needed to be and we much better understand the actual reasons we need it now which is to stream state to snapshots. 
- [x] Enable `Persistable<T>` to be constructed directly off `StoreConnector`, **Why?** Using a StoreConnector in order to aid factory functions helps with our API.
- [x] Setup aggregate based delay buffer so that aggregates are delayed based on age (this will be configured based on chain_id finalization times) **Why?** eventually we will want to have event sourcing aggregates that correspond to the blockchains that have originated the events we are processing. The reason for this is that if one chain has a reorg we will want to treat these differently. Ideally we set the snapshot delays to be the finalization times of the chain that is represented by the aggregate. This makes the snapshot be able to be delayed by various amounts based on the configuration. We cannot yet change our eventlog retrospectively and it would add a whole lot of complexity to do this but we are ready to handle this. 
- [x] Store sequence number with aggregate in snapshot. **Why?** so when the inserts are flushed to disk the disk contains the last seq and aggregate_id that the state knew so we can synchronize them
- [x] Add one commitlog per aggregate_id when sequencing events. **Why?** so that we hold separate streams of events and can manipulate them separately whilst reconstructing aggregate bound snapshots separately.
- [x] When requesting events we need to request with a hashmap of `agid -> timestamps`. **Why?** because we are using multiple aggregates we need to ask for the timestamps for each aggregate.
- [x] Make handlers run strictly sequentially to prepare for easily attaching context to publishers (`self.notify_sync(ctx, evt)`) **Why?** This is a good practice in general as it will aid in auditing but may help by allowing us to set the context as we handle the EnclaveEvent - we will need to forward context between actors still (see TypedEvent<T>) but this will simplify how often we must do this.

#### Event system
- [x] Add net request framework to query nodes for events **Why?** to prepare for requesting for libp2p events from the network.
- [x] Investigate what is required for evm refactor for syncing
- [x] Refactor evm into EvmSystem **Why?** previously we had one read provider per contract per chain listening for events and had to coordinate between all the providers in order to know when historical events had finished syncing. Now we have one read provider per chain that reads logs and sends them up to be routed and then parsed before being managed by either sending them to the bus or to the sync actor. **but why?** We had to do some major refactoring to the evm component to organize sync, ensure blockchain events have the correct timestamps on them and ensure there is a clearer distinction between histrical and live event management - I have added to this by setting up a kind of event pipeline that is neat and extensible moving forward. Also this will help with #1020
- [x] Moved the localhost check in the evm reader to a config parsing check. **Why?** because it is a better design. There are questions about the wisdom of checking this this way as you might run a localhost local node but we can look at this later if it becomes a problem.  
- [x] Get tests pasing as before

#### Synchronization
- [x] Add sync scaffold process on startup - This is rough and does basically what was happening before. 
- [x] Create basic test for Sync doing what it is currenty doing - just to check

---

### Next PR
Event writing anf snapshotting
- [ ] force context on all events and snapshots
  - [ ] inject context into bus and persistence at the start of all sequential handlers or use TypedEvents for notified handlers.
- [ ] Ensure blockheight watermark is saved on each aggregate write
- [ ] Setup second block based subscription
- [ ] Query snapshot information
- [ ] Delay snapshot information by finalization time

Event system
- [ ] Delay live evm processing by two blocks

Synchronization
- [ ] Fetch initial information for Sync from eventsystem
- [ ] Fetch historical events from evm
- [ ] Fetch historical events from net
- [ ] Sort events by timestamp before publishing them
- [ ] Open request edgecase detect in complete events

General
- [ ] Test a sync in the middle of aggregation events

## Summary by CodeRabbit

* **New Features**
  * Added network synchronization for coordinating events across peers.
  * Implemented multi-chain EVM event handling with historical and live event support.
  * Added provider caching for efficient blockchain interactions.

* **Refactor**
  * Redesigned event architecture to support context tracking and causation chains.
  * Restructured EVM event processing pipeline for improved ordering and reliability.
  * Extended network protocol for peer-to-peer synchronization.


ryardley:
Made note of coderabbit stuff will address in future PRs

PR #1167, opened 2026-01-15
fix: require threshold_n nodes for committee finalization

hmzakhalid:
## Summary by CodeRabbit

* **Bug Fixes**
  * Increased finalization delay to reduce premature processing.
  * Corrected which committee threshold is used during finalization.

* **Chores**
  * Added prominent warning logs during network keypair auto-generation for clearer guidance.
  * Reduced log verbosity for peer-ID mismatch events to cut noise.
  * Updated deployed contract records (implementation addresses and block numbers).

### DIFF
--- a/crates/aggregator/src/committee_finalizer.rs
+++ b/crates/aggregator/src/committee_finalizer.rs
@@ -62,7 +62,7 @@ impl Handler<CommitteeRequested> for CommitteeFinalizer {
         let e3_id = msg.e3_id.clone();
         let submission_deadline = msg.submission_deadline;
 
-        const FINALIZATION_BUFFER_SECONDS: u64 = 5;
+        const FINALIZATION_BUFFER_SECONDS: u64 = 1;
 
         let e3_id_for_log = e3_id.clone();
         let fut = async move {


crates/entrypoint/src/net/keypair/generate.rs
```diff
@@ -29,6 +30,11 @@ pub async fn execute(config: &AppConfig) -> Result<PeerId> {
 pub async fn autonetkey(config: &AppConfig) -> Result<()> {
     let repositories = get_repositories(config)?;
     if !repositories.libp2p_keypair().has().await {
+        warn!("Auto-generating network keypair because 'autonetkey: true' is set and no keypair exists.");
+        warn!("This will create a NEW peer identity. If your data directory is not persistent");
+        warn!("(e.g., running in Docker without volumes), a new identity will be generated on each restart,");
+        warn!("which will cause network connectivity issues with other peers.");
+        warn!("For production use, run 'enclave net keypair (generate)/(set --net-keypair <YOUR_PEER_ID>)' once and ensure data persistence.");
```
ryardley: Do we really want these to be separate tracing events?

hmzakhalid: Sorry no! We can make them a single one.

ctrlc03 (dismissed):

Issue #1184, opened 2026-01-21
Fix smudging noise in trbfv_Pk


zahrajavar:
Change `e_sm: Polynomial<N>` to `e_sm: [Polynomial<N>; L]` and fix the corresponding commitment.

PR #1196, opened 2026-01-22
fix: commitments [skip-line-limit]

0xjei:
see #1184

## Summary by CodeRabbit

* **Refactor**
  * Public keys and secret components now use per-modulus (multi-polynomial) representation.
  * Commitment handling unified into a domain-separated, wrapper-driven API; challenge and verification flows updated to use it.
  * Per-modulus range and consistency checks added for clearer validation.

* **New Features**
  * Centralized commitment helpers exposed for reuse.
  * Ciphertext commitment generation switched to the new commitment API.

* **Chores**
  * On-chain verifier key constant updated.


0xjei:
we need https://github.com/gnosisguild/zkfhe-generator/pull/72 merged and check the CI again (I think the `greco` commitment is breaking things now)

PR #1211, opened 2026-01-27
feat: add e3-noir-prover crate for proof generation [skip-line-limit]

hmzakhalid:
Enable ciphernodes to generate proofs for cryptographic operations (key generation, decryption shares, etc.) using Noir circuits and the Barretenberg prover.

### High-Level Flow

```
1. Ciphernode receives CiphernodeSelected event
                    │
                    ▼
2. Generate BFV key pair (existing FHE logic)
                    │
                    ▼
3. Prepare circuit inputs (serialize to TOML)
                    │
                    ▼
4. Call NoirProver::generate_proof()
         │
         ├── Create work dir: ~/.enclave/noir/work/<e3_id>/
         │
         ├── Write inputs to ~/.enclave/noir/work/<e3_id>/inputs.toml
         │
         ├── Execute: bb prove --scheme ultra_honk \
         │              -b ~/.enclave/noir/circuits/pk_bfv.json \
         │              -w ~/.enclave/noir/work/<e3_id>/witness.gz \
         │              -o ~/.enclave/noir/work/<e3_id>/
         │
         └── Read proof from output file
                 │
                ▼
5. Broadcast KeyshareCreated { pubkey, proof_0, ... }
                 │
                ▼
6. Cleanup work directory (witness stays local, never transferred)
```

### Enclave Noir CLI

```
Noir prover management and proof generation

Usage: enclave noir [OPTIONS] <COMMAND>

Commands:
  status  Check Noir prover setup status (bb binary and circuits)
  setup   Download or update bb binary and circuit artifacts
  prove   Generate a proof for a circuit (for testing)
  verify  Verify a proof (for testing)
  help    Print this message or the help of the given subcommand(s)
```

### Directory Structure

```
~/.enclave/
└── noir/
    ├── version.json           # Tracks binary versions
    ├── bin/
    │   └── bb                  
    ├── circuits/               # Pre-compiled circuit artifacts
    │   ├── pk_bfv.json         # T0: BFV public key generation circuit
    │   ├── dec_share_trbfv.json # Decryption share circuit
    │   ├── verify_shares.json  # Share verification circuit
    │   └── vk/                 # Verification keys
    │       ├── pk_bfv.vk
    │       └── ...
    └── work/                   # Runtime working directory
        └── <e3_id>/            # Per-request working directory
            ├── inputs.toml     # Circuit inputs 
            ├── witness.gz      # Generated witness 
            └── proof           # Generated proof 
```

Remaining Tasks:
- [x] Pin exact bb version and add checksum verification
- [x] Set up circuit artifact hosting on GitHub Releases and update download URLs
- [x] Implement T0 proof generation
  - [x] Map BFV key generation outputs to circuit inputs
  - [x] Generate Proof 0 when creating BFV key pair
  - [x] Create/cleanup `~/.enclave/noir/work/<e3_id>/`
  - [x] Include Proof 0 in KeyshareCreated message
  - [x] Verify received Proof 0 before accepting pubkey
  - [x] E2E tests with T0 proof generation and verification
- [ ] Bundle bb binary in Docker image for production deployments
- [ ] Add benchmarks for proof generation time and memory usage
- [x] E2E tests with multiple ciphernodes exchanging proofs

## Summary by CodeRabbit

* **New Features**
  * Full ZK prover added: CLI noir commands (status/setup), prover backend, actors for proof generation & verification, and optional ZK integration in node builder/runtime.
* **Bug Fixes**
  * Encryption key flow updated to emit pending/received events and verify ZK proofs before finalizing keys.
* **Chores**
  * Workspace manifests, CI, startup scripts and templates updated to provision the ZK prover.
* **Tests**
  * Extensive unit, integration, local e2e and backend tests for prover, witness, and verification flows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3811,7 +3811,7 @@ dependencies = [
 
 [[package]]
 name = "e3-zk-prover"
-version = "0.1.9"
+version = "0.1.11"
 dependencies = [
  "acir",
  "actix",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,6 +104,7 @@ e3-tests = { version = "0.1.11", path = "./crates/tests" }
 e3-trbfv = { version = "0.1.11", path = "./crates/trbfv" }
 e3-utils = { version = "0.1.11", path = "./crates/utils" }
 e3-safe = { version = "0.1.11", path = "./crates/safe" }
+e3-zk-prover = { version = "0.1.11", path = "./crates/zk-prover" }
 e3-zk-helpers = { version = "0.1.11", path = "./crates/zk-helpers" }
 e3-parity-matrix = { version = "0.1.11", path = "./crates/parity-matrix" }
 
--- a/crates/zk-prover/src/actors/zk_actor.rs
+++ b/crates/zk-prover/src/actors/zk_actor.rs
@@ -37,7 +37,10 @@ impl Handler<ZkVerificationRequest> for ZkActor {
     type Result = ();
 
     fn handle(&mut self, msg: ZkVerificationRequest, _ctx: &mut Self::Context) -> Self::Result {
-        debug!("Verifying proof for circuit: {}", msg.proof.circuit);
+        debug!(
+            "Verifying proof for circuit: {} (party {})",
+            msg.proof.circuit, msg.key.party_id
+        );
 
         let e3_id_str = msg.e3_id.to_string();
         let result = self.prover.verify_proof(
@@ -45,6 +48,7 @@ impl Handler<ZkVerificationRequest> for ZkActor {
             &msg.proof.data,
             &msg.proof.public_signals,
             &e3_id_str,
+            msg.key.party_id,
         );
 
         let response = match result {
--- a/crates/zk-prover/src/prover.rs
+++ b/crates/zk-prover/src/prover.rs
@@ -122,8 +122,14 @@ impl ZkProver {
         ))
     }
 
-    pub fn verify(&self, proof: &Proof, e3_id: &str) -> Result<bool, ZkError> {
-        self.verify_proof(proof.circuit, &proof.data, &proof.public_signals, e3_id)
+    pub fn verify(&self, proof: &Proof, e3_id: &str, party_id: u64) -> Result<bool, ZkError> {
+        self.verify_proof(
+            proof.circuit,
+            &proof.data,
+            &proof.public_signals,
+            e3_id,
+            party_id,
+        )
     }
 
     pub fn verify_proof(
@@ -132,6 +138,7 @@ impl ZkProver {
         proof_data: &[u8],
         public_signals: &[u8],
         e3_id: &str,
+        party_id: u64,
     ) -> Result<bool, ZkError> {
         if !self.bb_binary.exists() {
             return Err(ZkError::BbNotInstalled);
@@ -148,17 +155,20 @@ impl ZkProver {
             )));
         }
 
+        let verification_subdir = format!("verify_party_{}", party_id);
+
         debug!(
-            "verifying proof for circuit {} using VK: {}",
+            "verifying proof for circuit {} (party {}) using VK: {}",
             circuit.as_str(),
+            party_id,
             vk_path.display()
         );
 
-        let job_dir = self.work_dir.join(e3_id);
+        let job_dir = self.work_dir.join(e3_id).join(&verification_subdir);
         let out_dir = job_dir.join("out");
         fs::create_dir_all(&out_dir)?;
 
-        let proof_path = job_dir.join("proof_to_verify");
+        let proof_path = job_dir.join("proof");
         let public_inputs_path = out_dir.join("public_inputs");
 
         fs::write(&proof_path, proof_data)?;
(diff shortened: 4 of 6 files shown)


crates/keyshare/src/threshold_keyshare.rs
```diff
@@ -408,28 +437,33 @@ impl ThresholdKeyshare {
...
+                }
+                TrBFVResponse::CalculateDecryptionKey(_) => {
+                    self.handle_calculate_decryption_key_response(msg)
+                }
+                TrBFVResponse::CalculateDecryptionShare(_) => {
+                    self.handle_calculate_decryption_share_response(msg)
+                }
+                _ => Ok(()),
+            },
+            ComputeResponseKind::Zk(zk) => match zk {
```
ryardley: 👍🏾 +1

ryardley (commented):
Have you considered the approach of creating an actor that listens to the event bus and converts DecryptionShareCreated events or other proof carrying events to VerifiedDecryptionShareCreated events and only collect the VerifiedDecryptionShareCreated events? 
I think this will make life easier as it keeps our dependencies isolated is easy to test and makes everything much more syncable. Just trying to suggest to think in pipelines.

crates/keyshare/src/encryption_key_collector.rs
```diff
@@ -118,6 +126,50 @@ impl Handler<EncryptionKeyCreated> for EncryptionKeyCollector {
...
+                warn!(
+                    "External key from party {} is missing T0 proof - rejecting",
+                    pid
+                );
+                return;
+            };
+
+            // Verify proof if ZK backend is available
+            if let Some(ref backend) = self.zk_backend {
+                let prover = ZkProver::new(backend);
```
ryardley: If we are going to verify in core we might want to consider a prover to prove and a verifier to verify - it makes it easier to define rules like only verifiers are allowed to run outside synchronous code. We could do things like make it so verifiers are clonable but provers are not. etc.

crates/keyshare/src/encryption_key_collector.rs
```diff
@@ -65,17 +66,24 @@ pub struct EncryptionKeyCollector {
...
+    zk_backend: Option<ZkBackend>,
 }
 
 impl EncryptionKeyCollector {
-    pub fn setup(parent: Addr<ThresholdKeyshare>, total: u64, e3_id: E3id) -> Addr<Self> {
+    pub fn setup(
+        parent: Addr<ThresholdKeyshare>,
+        total: u64,
+        e3_id: E3id,
+        zk_backend: Option<ZkBackend>,
```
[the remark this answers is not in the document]
ryardley: Thinking through this we should not have an IO bound dependency in core at all. It will make testing much easier if we use some kind of indirection such as a `Box<dyn Verifier>` (which we have not really used yet) or use events which would be more idiomatic within the codebase - leading towards the idea of verifying in stream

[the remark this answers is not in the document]
ryardley: I think it might also be fine to have an actor that acts as the verifier that you send the proof to then we just hold a Recipient<VerifyProof> the point is to avoid IO dependency in core code but then you have async routines in which case it makes more sense to do stream verification.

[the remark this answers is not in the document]
hmzakhalid: I like the event pipeline idea. What do you think about a dedicated ZkActor that centralizes all ZK operations?
Flow would be:
```
ThresholdKeyshare generates keys
  → publishes EncryptionKeyPending { pk_bfv, params }
  
ZkActor receives EncryptionKeyPending
  → sends ComputeRequest to Multithread for proof generation
  → receives ComputeResponse with proof
  → publishes EncryptionKeyCreated (with proof attached)

For external keys:
Network → EncryptionKeyReceived → ZkActor verifies → EncryptionKeyCreated

EncryptionKeyCollector only subscribes to EncryptionKeyCreated (already verified)
```

Would also add a ZkVerifier type for the lightweight verification path as you suggested.
Does this align with what you had in mind? Happy to refactor this way if so.

crates/events/src/enclave_event/proof.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// Encrypted shares proof (T2/T3).
+    EncShares,
+    /// Decryption share proof (T4/T5).
+    DecShares,
+    /// Public key aggregation proof (T6).
+    PkAgg,
+}
+
+impl CircuitName {
+    /// Get the file name for this circuit.
```
ryardley: This type should not know about filenames perhaps it might be better to call it a slug or give it a different semantic meaning or just delete this comment.

crates/zk-prover/src/actor.rs
```diff
@@ -0,0 +1,219 @@
...
+                external: false,
+            }) {
+                error!("Failed to publish EncryptionKeyCreated: {err}");
+            }
+            return;
+        }
+
+        let correlation_id = CorrelationId::new();
+        self.pending.insert(
+            correlation_id.clone(),
```
ryardley: `CorrelationId` is Copy so you probably don't need to clone.

PR #1229, opened 2026-01-29
chore: rename noir to zk prover and add zk ext for cn builder

hmzakhalid:


crates/zk-prover/src/prover.rs
```diff
@@ -92,48 +113,57 @@ impl NoirProver {
...
         if !output.status.success() {
             let stderr = String::from_utf8_lossy(&output.stderr);
-            return Err(NoirProverError::ProveFailed(stderr.to_string()));
+            return Err(ZkError::ProveFailed(stderr.to_string()));
         }
 
-        let proof = fs::read(&proof_path).await?;
-        info!("generated proof ({} bytes) for {}", proof.len(), e3_id);
+        let proof_data = fs::read(&proof_path).await?;
+        let public_output = fs::read(&public_inputs_path).await?;
```
ctrlc03: should we maybe generalize to public signals? it will contain both public inputs and public outputs

ctrlc03 (approved):

PR #1257, opened 2026-02-03
refactor: remove activate step [skip-line-limit]

### LINKED ISSUE #1226: Remove activate step
with moving publish input to program we can also remove activation. This step was only needed to ensure we had a ciphernode key but now with the lyfecycle contracts we have fixed stages. E3s will start accepting inputs from input start window until input deadline.

ctrlc03:
fix #1226


hmzakhalid (approved):
Amazing Work, Looks great.
The only thing now left is to figure out a way to make the user input window start from after the KeyPublished stage.

PR #1290, opened 2026-02-10
feat: add sk_share and e_sm_share

ctrlc03:
Add sk_share_computation and e_sm_share computation circuits into zk prover, as well as tests. 

## Release Notes

* **New Features**
  * Added support for two new zero-knowledge proof circuits for share computation, enabling proof generation for both secret key and smudging noise types.
  * Enhanced circuit selection logic to support multiple valid circuits per operation.

* **Tests**
  * Expanded end-to-end test coverage with commitment consistency tests for new share computation proof types.
  * Added test utilities for extracting field values from proof outputs.

### DIFF
--- a/crates/zk-helpers/src/circuits/dkg/share_computation/computation.rs
+++ b/crates/zk-helpers/src/circuits/dkg/share_computation/computation.rs
@@ -16,10 +16,10 @@ use crate::circuits::commitments::{
 use crate::computation::DkgInputType;
 use crate::dkg::share_computation::ShareComputationCircuit;
 use crate::dkg::share_computation::ShareComputationCircuitData;
+use crate::poly_coefficients_to_toml_json;
 use crate::CircuitsErrors;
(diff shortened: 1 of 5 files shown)


cedoor (dismissed):

crates/zk-helpers/src/circuits/dkg/share_computation/computation.rs
```diff
@@ -93,6 +94,8 @@ pub struct Inputs {
     pub y: Vec<Vec<Vec<BigInt>>>,
     /// Expected secret commitment (matches C1's compute_secret_commitment).
     pub expected_secret_commitment: BigInt,
+    /// Which secret type this witness is for (determines which circuit to run).
+    pub dkg_input_type: DkgInputType,
```
0xjei: I don't like this but we can refactor later by changing the signature of the `to_json` to optionally accept the input type

PR #1339, opened 2026-02-17
fix: libp2p usage patterns in e3-net crate

hmzakhalid:
Did a review of the net crate and found several areas where we could improve our libp2p usage and overall code quality

Key changes:
- Replaced unbounded HashSet with BloomFilter for event deduplication
- Fixed Kademlia to use our own protocol namespace instead of IPFS
- Removed manual DNS resolution (libp2p handles this)
- Added proper bootstrap triggering on first peer connection
- Refactored correlator to use typed keys instead of strings
- Fixed event loop to actually exit on shutdown
- Renamed Cid → ContentHash to avoid confusion with IPFS standard

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved peer failure detection and eviction for more reliable connections.
  * More resilient network ops with guarded error handling and clearer logs.

* **Refactor**
  * Event deduplication switched to a memory‑efficient probabilistic filter for lower memory use.
  * Dialing flow simplified (no DNS-based resolution) to reduce connection surprises.

* **Chores**
  * Network topic renamed to "enclave-gossip" and dependency set updated.

### DIFF
--- a/crates/evm/src/fix_historical_order.rs
+++ b/crates/evm/src/fix_historical_order.rs
@@ -22,7 +22,7 @@ impl FixHistoricalOrder {
         Self {
             dest: dest.into(),
             pending_sync_complete: None,
-            seen_ids: BloomFilter::with_rate(0.001, 10_000),
+            seen_ids: BloomFilter::with_rate(0.001, 10_000_000),
         }
     }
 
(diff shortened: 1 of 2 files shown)


ryardley (approved):
looks good!

PR #1343, opened 2026-02-17
fix: prevent SyncEnded dedup that blocks live EVM events after restart

hmzakhalid:
On restart, the sync process replayed persisted infrastructure events (`SyncEnded`, `EffectsEnabled`) from the EventStore back through the EventBus. Since these events have no fields, they always produce the same `EventId` hash, poisoning the bloom filter and causing the freshly published `SyncEnded` to be silently dropped. With the `EvmChainGateway` never receiving `SyncEnded`, it remained stuck in `BufferUntilLive` and all live EVM events were silently buffered rather than dispatched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Infrastructure events no longer trigger deduplication issues during event replay, ensuring proper re-publication.

* **Refactor**
  * Simplified network interface shutdown handling.
  * Enhanced sync event tracking with correlation IDs for improved traceability.

* **Tests**
  * Added tests validating infrastructure event detection and filtering during replay.

### DIFF
--- a/crates/events/src/enclave_event/enable_effects.rs
+++ b/crates/events/src/enclave_event/enable_effects.rs
@@ -4,18 +4,23 @@
 // without even the implied warranty of MERCHANTABILITY
 // or FITNESS FOR A PARTICULAR PURPOSE.
 
+use crate::CorrelationId;
 use actix::Message;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


ryardley (approved):
Nice one!

PR #1362, opened 2026-02-25
feat: integrate proof of correct share decryption [skip-line-limit]

### LINKED ISSUE #1361: Integrate c4 circuits into ciphernodes

ctrlc03:
fix #1361 

## Summary by CodeRabbit

* **New Features**
  * Multi-stage verification for threshold shares, including batch proof verification and DKG decryption proofs.
  * Publishable per-party decryption-key sharing event/document with attached C4 proofs.
  * Threshold share messages now include per-party signed proof payloads for richer verification.
  * New accessors to read encrypted-share ciphertext bytes and modulus count.

* **Tests**
  * Adjusted integration test timing parameter.

### DIFF
--- a/crates/keyshare/src/threshold_keyshare.rs
+++ b/crates/keyshare/src/threshold_keyshare.rs
@@ -12,13 +12,14 @@ use e3_events::{
     prelude::*, trap, BusHandle, CiphernodeSelected, CiphertextOutputPublished, ComputeRequest,
     ComputeResponse, ComputeResponseKind, CorrelationId, DecryptionKeyShared,
     DecryptionshareCreated, Die, DkgProofSigned, DkgShareDecryptionProofRequest,
-    DkgShareDecryptionProofResponse, E3RequestComplete, E3id, EType, EnclaveEvent,
-    EnclaveEventData, EncryptionKey, EncryptionKeyCollectionFailed, EncryptionKeyCreated,
-    EncryptionKeyPending, EventContext, KeyshareCreated, PartyId, PartyProofsToVerify,
-    PkGenerationProofRequest, PkGenerationProofSigned, Proof, ProofType, Sequenced,
-    ShareComputationProofRequest, ShareEncryptionProofRequest, SignedProofPayload, ThresholdShare,
-    ThresholdShareCollectionFailed, ThresholdShareCreated, ThresholdSharePending, TypedEvent,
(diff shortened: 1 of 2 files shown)


crates/keyshare/src/threshold_keyshare.rs
```diff
@@ -1001,13 +1043,177 @@ impl ThresholdKeyshare {
...
         let (msg, ec) = msg.into_components();
         info!("AllThresholdSharesCollected");
+        let state = self.state.try_get()?;
+        let e3_id = state.get_e3_id();
+        let own_party_id = state.party_id;
+
+        // Build verification requests for other parties' proofs
+        let mut party_proofs_to_verify: Vec<PartyProofsToVerify> = Vec::new();
+        let mut no_proof_parties: HashSet<u64> = HashSet::new();
+        for (share, proofs) in msg.shares.iter().zip(msg.share_proofs.iter()) {
```
hmzakhalid: All proofs from a party are flattened into a single `Vec<SignedProofPayload>` and passed to the verifier,so the verifier only checks the proof that are present, we should add a check here if there's any missing proof and that the expected set is complete

A malicious sender can omit the proofs that would fail and still be treated as `all_verified: true`.

crates/keyshare/src/threshold_keyshare.rs
```diff
@@ -1001,13 +1043,177 @@ impl ThresholdKeyshare {
...
+
+            if honest_count < threshold {
+                warn!(
+                    "Too few honest parties for E3 {} ({} honest < {} threshold) — cannot proceed",
+                    e3_id, honest_count, threshold
+                );
+                // TODO: emit E3Failed event
+                self.pending_verification_shares = None;
+                return Ok(());
+            }
```
hmzakhalid: Should emit E3Failed so it doesnt hang, the on chain E3 would timeout and the requestor can claim refund

crates/zk-prover/src/actors/proof_request.rs
```diff
@@ -435,16 +441,34 @@ impl ProofRequestActor {
             return;
         };
 
+        // Sign C3a proofs (SkShareEncryption) — keyed by (recipient, row)
+        let mut signed_c3a_map: HashMap<usize, Vec<SignedProofPayload>> = HashMap::new();
```
hmzakhalid: We should use a BTreeMap here instead so the ordering is deterministic

PR #1444, opened 2026-03-17
feat: rebroadcast for dkg and decryption

ctrlc03:
ensure that dkg and threshold events are forwarded even when we have network errors - this caused issues in dkg where certain nodes would not receive keyshares from others

## Summary by CodeRabbit

* **New Features**
  * Document notifications are now periodically rebroadcasted during ephemeral enclave phases, improving delivery reliability during extended processing periods.


ryardley (commented):
Just quick first impressions here.

The concept of "Document Publishing" should be a generic thing ideally and should not in principle have anything to do with E3Stage or know about DKG at all so merging these concepts together in an actor is probably not quite right.

I will have a bit more of a look later at this tomorrow when I have a bit more time.

PR #1460, opened 2026-03-19
feat: connect c1 and c5 [skip-line-length]

### LINKED ISSUE #1452: Cross-circuit check: C1→C5 — verify C5 uses C1 pk_commitments of honest parties
Ensure the C5 (pk aggregation) proof is built against the actual C1 pk commitments of the parties that passed C1 verification. Today the aggregator filters dishonest parties after C1 and then builds C5 from the honest set; we need an explicit check that the commitments in C5’s inputs (or public signals) match the C1 pk_commitment values of those parties.

**Tasks:**
- In `PublicKeyAggregator` (`crates/aggregator/src/publickey_aggregator.rs`), after C1 verification completes and honest parties are determined:
  - From each honest party’s C1 proof (already available in the VerifyingC1 state), extract the pk_commitment (e.g. second-to-last field in public signals; see tests in `zk-prover/tests/local_e2e_tests.rs`).
  - When building or validating the C5 proof request/inputs: ensure the list of `expected_threshold_pk_commitments` passed to C5 matches, in the same order, the C1 pk_commitments of the honest parties.
  - If C5 proof is received from another component (zk-prover): after receiving the signed C5 proof, parse its public signals and compare `expected_threshold_pk_commitments` (per Issue 1) to the cached C1 pk_commitments; on mismatch, treat as verification failure (and do not publish PublicKeyAggregated or mark appropriately).
- Add a test or assertion that a C5 proof built with a wrong commitment is rejected or the aggregator does not complete.

**Acceptance criteria:**
- C5 is only accepted when its expected threshold pk commitments match the C1 pk_commitments of the honest parties in the same order.
- No new event types; reuse existing aggregator failure/validation paths.

ctrlc03:
fix #1452 

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened verification of commitment consistency across parties during aggregation
  * Improved detection and handling of faulty parties with automatic re-aggregation of valid participants
  * Enhanced early validation to reject invalid requests and prevent downstream processing errors

* **Chores**
  * Adjusted test timeout configurations

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,6 +12,8 @@
 !pnpm-lock.yaml
 !rust-toolchain.toml
 !**/*.sh
+!flake.nix
+!flake.lock
 
 # Explicitly ignore build artifacts everywhere
 target/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -463,6 +463,20 @@ jobs:
             target/debug/pack_e3_params
           if-no-files-found: error
 
+  check-nix-flake:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: cachix/install-nix-action@v30
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+        # if the following fails it means some of our git dependencies or bb has updated
+        # to fix this basically what you need to do here is copy the actual hashes
+        # into the flake.nix where those hashes are expected. if you want to do this locally
+        # install the nix package manager and try running the scripts below.
+      - run: bash ./scripts/nix-versions.sh # display bad versions when deps change
+
   ciphernode_integration_test:
     needs: [detect_changes, integration_prebuild, build_enclave_cli, build_sdk]
     if: needs.detect_changes.outputs.ciphernode_e2e == 'true'
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -138,6 +138,28 @@ jobs:
           cache-to: |
             type=gha,mode=max,scope=e3-support
 
+  build-nix-flake:
+    name: Build Nix Flake
+    runs-on: ubuntu-latest
+    needs: validate-and-prepare
+    steps:
+      - uses: actions/checkout@v6
+      - uses: cachix/install-nix-action@v30
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+            - uses: cachix/install-nix-action@v4
+      - uses: cachix/cachix-action@v15
+        with:
+          name: enclave
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+        # if the following fails it means some of our git dependencies or bb has updated
+        # to fix this basically what you need to do here is copy the actual hashes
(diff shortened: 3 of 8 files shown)


crates/aggregator/src/publickey_aggregator.rs
```diff
@@ -25,6 +25,34 @@ use std::collections::{BTreeSet, HashMap};
...
+}
+
+/// Convert `Vec<Option<ArcBytes>>` to `Vec<ArcBytes>`, substituting 32 zero bytes
+/// for any `None` entry. The C5 prover will detect these as commitment mismatches.
+fn unwrap_c1_commitments(commitments: &[Option<ArcBytes>]) -> Vec<ArcBytes> {
+    commitments
+        .iter()
+        .map(|opt| {
+            opt.clone()
+                .unwrap_or_else(|| ArcBytes::from_bytes(&[0u8; 32]))
```
cedoor: this will fail anyway, so better to throw an error here if it's not unwrappable

crates/aggregator/src/publickey_aggregator.rs
```diff
@@ -329,14 +399,16 @@ impl PublicKeyAggregator {
         self.state.try_mutate(&ec, |_| {
             Ok(PublicKeyAggregatorState::GeneratingC5Proof {
                 public_key: pubkey.clone(),
-                keyshare_bytes,
+                keyshare_bytes: honest_keyshares,
```
hmzakhalid: shouldnt this stay as `keyshare_bytes`?

crates/aggregator/src/publickey_aggregator.rs
```diff
@@ -692,6 +769,64 @@ impl PublicKeyAggregator {
...
+                            SignedProofFailed {
+                                e3_id: self.e3_id.clone(),
+                                faulting_node,
+                                proof_type: ProofType::C1PkGeneration,
+                                signed_payload: signed_payload.clone(),
+                            },
+                            ec.clone(),
+                        ) {
+                            error!("Failed to publish SignedProofFailed for C1 mismatch at index {}: {err}", idx);
+                        }
```
hmzakhalid: There should be a state transition here but we currently dont have any existing failure state. perhaps we should add a failure state here

crates/zk-helpers/src/circuits/threshold/pk_generation/computation.rs
```diff
@@ -130,12 +130,19 @@ impl Computation for Bits {
...
         let e_sm_bit = calculate_bit_width(BigInt::from(data.e_sm_bound.clone()));
-        let pk_bit = calculate_bit_width(BigInt::from(data.pk_bound.clone()));
+
+        let (threshold_params, _) =
+            build_pair_for_preset(preset).map_err(|e| CircuitsErrors::Other(e.to_string()))?;
+        let mut pk_bit = 0;
+        for &qi in threshold_params.moduli() {
+            let bound = BigUint::from(qi - 1);
+            pk_bit = pk_bit.max(calculate_bit_width(BigInt::from(bound)));
+        }
```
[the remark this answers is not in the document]
cedoor: everything is centered now, so we need to compute the bound with `(max(qi) - 1) / 2` in c1 too.

PR #1526, opened 2026-04-27
fix: config circuit [skip-line-limit]

0xjei:
This PR fixes the [config](https://github.com/gnosisguild/enclave/blob/main/circuits/bin/config/src/main.nr) circuit wrong `esm` and `r1_bounds` computations due to being not aligned with latest changes for the new parameter set (lambda = 60) and other fixes. Note that this does not change anything in terms of parameter generation or configs (just that the circuit was doing the wrong checks).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected verification parameter bounds formulas for threshold cryptographic configuration.
  * Adjusted parameter search configuration values for optimized threshold parameter selection.

* **Documentation**
  * Clarified share decryption bounds formula documentation for improved readability.


cedoor (approved):
lgtm

PR #1529, opened 2026-05-03
feat: stop encrypting DKG shares to self

hmzakhalid:
During threshold key generation, every node was previously encrypting a share of its own secret to itself, sending it over the wire, then decrypting it back, a wasteful round-trip with no security benefit (a node already knows its own share). This PR removes the self-loop end-to-end.

## Summary by CodeRabbit

* **Refactor**
  * DKG/share flow reorganized to treat the node’s own plaintext share separately and integrate it into aggregation.
* **New Features**
  * Sparse recipient encryption (ability to skip self slot) with aligned witness handling and APIs to support sparse recipient sets.
* **Bug Fixes**
  * Improved validation and error handling for share/plaintext shapes, dimensions, and proof set completeness.
* **Documentation**
  * Updated DKG encryption, proof-indexing, and recursive aggregation descriptions.
* **Tests**
  * Added/updated tests covering sparse skip-index behavior and related folding assertions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnosisguild/enclave/pull/1529)


0xjei (approved):
utACK - looks good as far as CI passes. Let's wait for @zahrajavar review!

zahrajavar (approved):

PR #1535, opened 2026-05-06
fix(zk-prover): node proof aggregation stalls

### LINKED ISSUE #1534: HLT-005: Proof-combining step can drop inputs and stall DKG completion
The step that combines node-level proofs before committee publication can stall if proof pieces arrive slightly out of order or if the fold job fails. In the bad case, the round stops making progress without emitting a clear failure signa

hmzakhalid:
closes #1534


0xjei (approved):
utACK! very neat, I was experiencing this sometimes during my runs for benches, so amazing work!

0xjei:
@coderabbitai review

PR #1537, opened 2026-05-06
fix: fail pk and plaintext ag err path [skip-line-limit]

### LINKED ISSUE #1536: HLT-006 / HLT-007: make PK and plaintext aggregation fail the round explicitly instead of stalling
Two late-stage aggregation paths can hit fatal runtime errors without converting the E3 into a clear failed state. This causes the round to stop making progress while the system still looks active from the outside.

There are two related failure modes here:

1. Public key aggregation can hit a fatal compute or validation error and stop locally without driving the round into a terminal failure state.
2. Plaintext / decryption aggregation can do the same thing later in the flow, especially when required shares, proofs, or aggregation inputs are missing or invalid.

In both cases, the practical outcome is the same:
- the node knows the round is broken
- the protocol does not always publish a clear failure signal
- the E3 can appear to hang until some later timeout or manual intervention

hmzakhalid:
close #1536 

## Summary by CodeRabbit

* **New Features**
  * Added explicit terminal failure event (E3Failed) and enriched aggregation outputs to include full public keys, commitments, per-node proof bundles, and decryption-aggregation proofs.

* **Bug Fixes**
  * Stronger error handling: compute/aggregation errors now route to terminal failures, normalize internal state, and prevent invalid continuation; mixed/missing proof cases treated as failures.

* **Documentation**
  * Updated protocol flow-trace docs with refreshed DKG/decryption workflows and failure behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnosisguild/enclave/pull/1537)


0xjei (dismissed):
utACK!

0xjei (approved):

ctrlc03 (approved):

PR #1550, opened 2026-05-21
feat: enable full e2e proof agg and slashing [skip-line-limit]

### LINKED ISSUE #1549: Set-level binding different from per-party attribution
The circuit enforces `committee_hash == keccak256(committee_members[0..N])`, which proves the proof
was produced _for this committee as a set_. It does not prove that the node-fold proof at row `i`
was produced by the keypair that registered as `topNodes[party_ids[i]]`. The `party_ids → address`
mapping is asserted only by off-chain convention: the aggregator orders entries by submitted
`party_id` and the circuit trusts that this index aligns with `topNodes[i]`.

A malicious aggregator holding valid fold proofs from every committee member could reassign which
`party_id` each proof is attributed to and produce a valid aggregated proof without any individual
operator having consented to the attribution. With H = N this is "merely" an attribution gap; once H
< N (post-expulsion), the aggregator also gains free choice over which subset to include.

This matters for slashing: identifying that a specific fold proof is malformed is not enough — there
is no cryptographic chain from "fold proof at row `i` is bad" to "address `topNodes[i]` produced
it."

The original design had a mechanism that closes this gap:

- Each node signs `sk_agg_commit_i` (or equivalent) with their registered key and shares it with the
  aggregator.
- On-chain verification:
  `signer = ecrecover(sk_agg_commits[i], signatures[i]); require(signer == registry[party_ids[i]].address)`

This creates a complete attribution chain: the entity that registered as party `i` is the same
entity that produced and signed off on the material at row `i`. I'd suggest this warrants its own
follow-up issue if not intended for this PR, so it doesn't get lost.

0xjei:
Several reviewers (and review LLMs) have flagged a concern that the committee-hash binding is set-level rather than per-party, and that "there is no cryptographic chain from 'fold proof at row i is bad' to 'address topNodes[i] produced it.'"

That chain *is* on this branch, in the Phase B fold-attestation work. For each honest party, `DkgFoldAttestationVerifier` enforces on-chain:

1. `isCommitteeMemberActive(e3Id, node)` for the claimed operator at that `partyId`.
2. `ecrecover(EIP-712 digest of (chainId, e3Id, partyId, skAggCommit, esmAggCommit), signature) == node`.
3. The signed `skAggCommit` / `esmAggCommit` equal the proof's public inputs at that party's slot.

Plus the bundle must contain exactly `h` attestations (derived from the proof shape) in strict ascending `partyId` order — so an aggregator holding fold proofs from all members can neither reassign attribution nor pick an H-subset.

Per-party commitments are persisted to `dkgPartyIds`, `dkgSkAggCommits`, `dkgEsmAggCommits` so downstream slashing can attribute a bad fold at row `i` to `topNodes[dkgPartyIds[e3Id][i]]`.

**Files to audit:**
- `packages/enclave-contracts/contracts/verifiers/DkgFoldAttestationVerifier.sol`
- `packages/enclave-contracts/contracts/lib/DkgFoldAttestationLib.sol`
- `_verifyAndStoreDkgAnchors` in `packages/enclave-contracts/contracts/registry/CiphernodeRegistryOwnable.sol`

**Docs:**
- `docs/pages/cryptography.mdx` — "Per-node attestations and topNodes binding" subsection
- `docs/pages/internals/dkg.mdx` — "Proof Aggregation and On-Chain Anchoring" section

## Summary by CodeRabbit

* **New Features**
  * On-chain proof aggregation flow with per-node fold-attestation bundles and EIP‑712 attestations.
  * Configurable proof-aggregation mode and multithread benchmarking (CLI/env flags).
  * Integration: CRISP dev/deploy helpers, env sync, and improved local deploy/test workflows.

* **Documentation**
  * Expanded docs on proof aggregation, fold-attestations, verifier syncing, and benchmarking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnosisguild/enclave/pull/1550?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ctrlc03 (approved):
utACK

PR #1597, opened 2026-06-12
fix: log flooding by redialing old entries

ctrlc03:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents repeated retries on stale peer identifiers; records mismatches, resets stale failure history, and cleans stale addresses.
  * Publishes corrected addresses and always attempts reconnect with the discovered peer identity; triggers bootstrap only on first mismatch.
  * Improved, clearer logging around peer ID mismatches and recovery behavior.

* **Tests**
  * Added integration and unit tests validating peer-ID stripping, single-mismatch handling, and successful recovery.

### DIFF
--- a/crates/net/src/dialer.rs
+++ b/crates/net/src/dialer.rs
@@ -131,17 +131,30 @@ async fn wait_for_connection(
                     } => {
                         trace!("OutgoingConnectionError!");
                         if connection_id == dial_connection {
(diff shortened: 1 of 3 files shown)


hmzakhalid (approved):
Thanks a lot Ale!
utACK

PR #1598, opened 2026-06-12
fix: update new parameter set for secure [skip-line-limit]

0xjei:


0xjei:
closing as we don't need to change the parameter set. I will reproduce the update of the search algo in PR #1596

PR #1599, opened 2026-06-13
feat: lock TTL

cristovaoth:
## Summary

Adds a lock sunset to `InterfoldToken`: a `NO_MORE_LOCKS` immutable timestamp after which all lock accounting is dead.

### Why
  
Lock accounting is per-transfer dead weight that never expires: every transfer from a once-locked account loops over its lock entries, loads their policies, and (while anything is still locked) makes an external `BONDING_REGISTRY.totalBonded` call, long after every curve has fully vested.

This PR gives that machinery a hard cutoff.

### DIFF
--- a/packages/interfold-contracts/contracts/token/InterfoldToken.sol
+++ b/packages/interfold-contracts/contracts/token/InterfoldToken.sol
@@ -921,11 +921,6 @@ contract InterfoldToken is
         return 0;
     }
 
-    /// @dev Earliest possible TGE timestamp.
-    function _earliestTge() internal view returns (uint256) {
-        return uint256(CCA_END) + TGE_COOLDOWN;
-    }
-
     /// @dev Ensures the policy cannot keep anything locked past the lock
     ///      sunset. Ending exactly at the sunset is allowed: the curve has
     ///      fully released and the hold has lapsed at that moment, which is
@@ -939,7 +934,7 @@ contract InterfoldToken is
             : curve.vestDuration;
 
         uint256 policyEnd = curve.anchor == Anchor.Tge
-            ? _earliestTge() + curveEnd
+            ? uint256(CCA_END) + TGE_COOLDOWN + curveEnd
             : curve.start + curveEnd;
 
         if (policyEnd > NO_MORE_LOCKS) {
--- a/packages/interfold-contracts/contracts/token/InterfoldToken.sol
+++ b/packages/interfold-contracts/contracts/token/InterfoldToken.sol
@@ -927,7 +927,9 @@ contract InterfoldToken is
     }
 
     /// @dev Ensures the policy cannot keep anything locked past the lock
-    ///      sunset.
+    ///      sunset. Ending exactly at the sunset is allowed: the curve has
+    ///      fully released and the hold has lapsed at that moment, which is
+    ///      also when the sunset takes effect.
     function _validatePolicyMaturity(LockPolicy calldata policy) internal view {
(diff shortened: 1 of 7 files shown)


packages/interfold-contracts/contracts/token/InterfoldToken.sol
```diff
@@ -575,17 +596,23 @@ contract InterfoldToken is
...
      */
     function _update(
         address from,
         address to,
         uint256 value
     ) internal override(ERC20, ERC20Votes) {
+        if (block.timestamp >= NO_MORE_LOCKS) {
+            super._update(from, to, value);
+            return;
+        }
```
hmzakhalid, line 614: the condition does not check whether `tge()` has actually been called. So public transfers can happen even though `tgeTimestamp == 0`, that violates the lifecycle model where public transfers open after TGE

packages/interfold-contracts/contracts/token/InterfoldToken.sol
```diff
@@ -894,6 +921,30 @@ contract InterfoldToken is
...
+        Curve calldata curve = policy.unlock;
+        // The curve validation guarantees cliff <= vest when vest != 0, so
+        // the curve fully releases at cliff (vest == 0) or at vest end.
+        uint256 curveEnd = curve.vestDuration == 0
+            ? curve.cliffDuration
+            : curve.vestDuration;
+
+        uint256 policyEnd = curve.anchor == Anchor.Tge
+            ? uint256(CCA_END) + TGE_COOLDOWN + curveEnd
+            : curve.start + curveEnd;
```
hmzakhalid, line 938: this is validating against the earliest possible TGE,  but the actual unlock math uses actual tgeTimestamp
i.e
```
earliest TGE = Jan 1
actual tge() call = Feb 1
policy = 24 month linear from TGE
```
the validator would then check
`Jan 1 + 24 months <= NO_MORE_LOCKS`

But the real policy ends:
`Feb 1 + 24 months`

So if `NO_MORE_LOCKS` is not buffered enough then the sunset can cut off the tail of the policy early

hmzakhalid (commented):
Thanks a lot for the PR. I like this direction, just left a few comments!

packages/interfold-contracts/contracts/token/InterfoldToken.sol
cristovaoth: Right, I see where you're coming from.

The idea for NO_MORE_LOCKS is an absolute sunset, set at deployment to a point in the future with an unequivocal margin of safety. There should be no ambiguity to the fact that at this absolute point in the future no other locks will remain.

And the idea is to make it a super cheap check (immutable + block timestamp ~20 gas) that acts as a lean short-circuit to ensure the contract eventually behaves like a standard ERC-20, regardless of other lifecycle or lock state variables.

packages/interfold-contracts/contracts/token/InterfoldToken.sol
cristovaoth: Yes, you're correct. This is the central trade-off of the sunset model: NO_MORE_LOCKS is the ultimate authority. If the TGE were delayed so significantly that a policy's tail pushed past that date, the sunset "wins" and the lock is truncated.

In practice, we mitigate this by configuring a conservative buffer at deployment, well beyond any realistic TGE delay.

Priority is an eventual gas-efficient exit to standard ERC-20 behavior over perfect policy enforcement in extreme-delay scenarios.

Issue #1690, opened 2026-07-11
Upgrade libp2p TLS and DNS stack to clear remaining RustSec advisories


hmzakhalid:
Audit reference: R01

## What can go wrong

After upgrading to libp2p 0.56.0, `cargo audit` still reports two advisories in `hickory-proto` 0.25.2: RUSTSEC-2026-0118 and RUSTSEC-2026-0119. The old `ring` 0.16 and `rustls-webpki` 0.101 advisory chain is gone.

## When it happens

Hickory is pulled in by `libp2p-dns` 0.44. The advisories concern DNSSEC NSEC3 closest-encloser processing and quadratic DNS message encoding. The current build does not enable Hickory's DNSSEC feature, which limits exposure to the first path, but both records correctly remain visible in the locked graph.

## What changed

The branch upgrades libp2p from 0.54.1 to 0.56.0, adapts its API changes, and validates QUIC, gossip, DHT, reconnect behavior, and mixed 0.54/0.56 peers. It does not force Hickory 0.26 into a dependency that requires the incompatible 0.25 line.

## Why the change works

The upgrade safely removes three old transport advisories, but it cannot honestly close this issue: RUSTSEC-2026-0119 is fixed only in Hickory 0.26.1+, and released libp2p 0.56 still pins the 0.25 family. Closure should wait for a stable compatible libp2p release or an upstream backport, followed by the same interoperability and soak tests.

## Checks

`e3-net` tests and strict Clippy pass, a four-node mixed-version QUIC/gossip/DHT drill passes, the full workspace suite passes, and the remaining audit result is exactly the two Hickory advisories above.

hmzakhalid:
Updated in [6032567751](https://github.com/theinterfold/interfold/commit/603256775197f8ef9ddf8edceb032ae01a09c4a8)

PR #1691, opened 2026-07-11
fix: harden ciphernode runtime and Rust architecture [skip-line-limit]

### LINKED ISSUE #1645: Bind C0 proofs to the canonical E3 request and committee slot
Audit reference: I01

## What can go wrong

A ciphernode could accept a valid C0 proof after it had been copied to a different E3 request, proof phase, committee slot, or public key.

## When it happens

This is checked when C0 key-generation proofs arrive during DKG, before the proof is allowed into the request's aggregation flow.

## What changed

The verifier now binds the E3 id, proof type, recovered signer, canonical party slot, proof bytes, and the public-key commitment recomputed from the submitted key.

## Why the change works

Every value that gives the proof meaning is checked together at the Rust verification boundary, so changing the request, owner, phase, or key makes the submission fail instead of being treated as the original proof.

## Checks

Regression tests cover wrong E3, phase, signer, slot, commitment, key bytes, and malformed keys.

### LINKED ISSUE #1646: Use the honest-party quorum for accusation voting
Audit reference: I02

## What can go wrong

Accusation voting counted the polynomial threshold T, even though the protocol requires the honest-party quorum H. That could resolve an accusation with the wrong number of votes.

## When it happens

The mismatch appears while the slashing actor tallies accusation votes or decides a vote at its timeout boundary.

## What changed

Voting and timeout decisions now derive and use the canonical H value from the committee parameters.

## Why the change works

Both paths now use the same quorum the protocol and circuits expect, including the exact boundary where a vote first becomes decisive.

## Checks

Canonical committee presets and exact-quorum boundary tests pass.

### LINKED ISSUE #1647: Fail closed on corrupt event logs and reconcile indexes in bounded pages
Audit reference: I03

## What can go wrong

A corrupt event record could be skipped, allowing the node to continue from incomplete history. Startup index repair could also load the whole log into memory.

## When it happens

Both problems occur during restart, when the event store reads durable history and reconciles its derived indexes before actors resume.

## What changed

Event-log reads now fail closed on corruption, and index reconciliation walks history in bounded pages.

## Why the change works

The node can no longer silently build state on a hole in its history, while a long but valid history uses a predictable amount of memory.

## Checks

Corrupt-tail, wrong-aggregate, validation-read, and paginated-reconciliation tests pass.

### LINKED ISSUE #1648: Keep network broadcast consumers alive after receiver lag
Audit reference: I04

## What can go wrong

A broadcast consumer treated temporary receiver lag as if the channel had closed permanently, so a busy node could silently stop consuming network events.

## When it happens

This happens when a network consumer falls behind the broadcast ring buffer during a burst but the producer is still healthy.

## What changed

Lag and channel closure are now handled separately: lag skips only the messages no longer retained and consumption continues.

## Why the change works

A recoverable backlog no longer kills the consumer; only an actual closed channel ends it.

## Checks

Tests cover continuation after lag and termination after a real channel close.

### LINKED ISSUE #1649: Make startup replay end-to-end backpressured
Audit reference: I05

## What can go wrong

A restart can leave thousands of post-snapshot events to replay. Reading all of them into one vector, or cloning them into subscriber mailboxes without waiting for capacity, can exhaust memory and put the node into a repeatable startup crash loop.

## When it happens

This is most likely after a long outage, stale snapshot, or recovery from a disk/RPC incident, when durable history is much larger than the in-memory EventBus and actor mailboxes.

## What changed

Replay now pages the EventStore in 1,024-event chunks, validates each aggregate's sequence continuity, and writes bounded sorted runs to temporary disk files. A bounded fan-in merge restores global timestamp order without retaining the whole backlog in memory. Each replayed event uses acknowledged subscriber fanout with a timeout, and EventStore query responses also wait for their recipient instead of bypassing mailbox capacity.

## Why the change works

Backpressure reaches from every subscriber all the way to the EventStore reader. A slow or stuck consumer therefore slows or fails startup explicitly instead of turning durable history into an unbounded in-memory backlog. Temporary spool files are removed on every exit path.

## Checks

A 5,120-event backlog, blocked-subscriber handling, saturated query recipients, sequence-gap rejection, external merge ordering, and spool cleanup are covered by focused tests.

### LINKED ISSUE #1650: Validate canonical H-of-N shapes before DKG aggregation
Audit reference: I06

## What can go wrong

DKG aggregation accepted proof-owner and committee-address lists with shapes that did not match the canonical H-of-N committee.

## When it happens

The check runs when the aggregator has collected key-generation proofs and is about to construct the DKG fold witness.

## What changed

Witness construction now requires exactly H unique, in-range proof owners and exactly N committee addresses in canonical order.

## Why the change works

The proof witness cannot be built from a duplicate, missing, foreign, or reordered committee member, so its dimensions stay aligned with the selected circuit preset.

## Checks

All committee presets and incomplete, duplicate, out-of-range, and wrong-N cases pass.

### LINKED ISSUE #1651: Match the SDK finalized-committee event to the contract ABI
Audit reference: I07

## What can go wrong

The Rust SDK subscribed to a finalized-committee event name that does not exist in the generated contract ABI, so applications could wait forever for an event the contract never emits.

## When it happens

This appears after sortition finalizes a committee and an SDK consumer is waiting to continue the E3 lifecycle.

## What changed

The SDK now subscribes to `SortitionCommitteeFinalized`, the event in the generated binding and contract.

## Why the change works

The listener and the deployed contract now agree on the same ABI event signature.

## Checks

The SDK ABI/event-name regression passes.

### LINKED ISSUE #1652: Bound ciphernode startup with a configurable readiness deadline
Audit reference: I08

## What can go wrong

Ciphernode startup could wait forever if a subsystem never reported ready, leaving a hung process that looked alive but could never serve work.

## When it happens

This occurs between process launch and readiness, for example when RPC, networking, storage recovery, or an actor stalls.

## What changed

Startup now has a validated, non-zero configurable deadline and returns a typed timeout error.

## Why the change works

The service manager gets a definite failure it can restart or alert on instead of an indefinitely half-started node.

## Checks

Default, zero-value rejection, successful readiness, and timeout tests pass.

### LINKED ISSUE #1653: Serialize pending nonce allocation for each chain signer
Audit reference: I09

## What can go wrong

Two concurrent EVM writers using the same signer and chain could read the same pending nonce and submit conflicting transactions.

## When it happens

The race occurs between reading the pending nonce and RPC acceptance whenever protocol or value-bearing writes are issued concurrently.

## What changed

Nonce lookup through transaction acceptance is serialized per chain and signer.

## Why the change works

Only one writer can reserve that signer's next nonce at a time, while unrelated chains or signers can still make progress independently.

## Checks

Concurrent nonce-allocation and EVM writer tests pass.

### LINKED ISSUE #1654: Replace probabilistic event deduplication and reject timestamp conflicts
Audit reference: I10

## What can go wrong

Bloom-filter false positives could discard a new event, and two different events with the same timestamp could be mistaken for an idempotent replay.

## When it happens

Deduplication runs as events enter the bus and store, including during restart replay and peer synchronization.

## What changed

The node now keeps bounded exact event ids and compares the complete event when a timestamp already exists.

## Why the change works

Exact ids remove probabilistic data loss, and full comparison distinguishes a harmless duplicate from conflicting durable history.

## Checks

Hash-collision, eviction, exact-duplicate, and conflicting-payload tests pass.

### LINKED ISSUE #1655: Redact and bound sensitive structured tracing fields
Audit reference: I11

## What can go wrong

Secrets or attacker-sized structured fields could enter retained logs and telemetry, leaking credentials or consuming excessive storage.

## When it happens

The exposure exists anywhere tracing records request data, inherited span fields, configuration, URLs, tokens, or malformed remote values.

## What changed

Sensitive field-name variants are redacted and structured names and values are bounded before retention; terminal rendering also neutralizes unsafe control data.

## Why the change works

Filtering happens at the telemetry boundary, so callers do not have to remember to sanitize every individual log statement.

## Checks

Sentinel redaction, inherited-span, ANSI, and truncation tests pass.

### LINKED ISSUE #1656: Bound historical peer synchronization and require cursor progress
Audit reference: I12

## What can go wrong

A peer could request an enormous history scan, keep returning a non-advancing cursor, or force too many concurrent synchronization jobs.

## When it happens

This is reachable when nodes reconnect and exchange historical events, especially when the remote peer is stale, buggy, or hostile.

## What changed

History sync now caps scans, concurrency, pages, events, bytes, and elapsed time, and requires every cursor to advance.

## Why the change works

Each peer request has a finite resource budget and a stalled cursor is stopped rather than looping over the same data.

## Checks

Limit, concurrency, timeout-release, cursor-progress, and total-budget tests pass.

### LINKED ISSUE #1657: Fail DAppNode startup closed on invalid credentials and false health
Audit reference: I13

## What can go wrong

The DAppNode launch path exposed credentials through environment variables, continued after provisioning failures, and treated a surviving PID as proof that the service was healthy.

## When it happens

These failures happen during container provisioning, restart, or health polling before the node is ready to participate.

## What changed

Credentials now use restricted files, inputs and CLI calls fail closed, plaintext setup material is removed, and health checks verify the local service rather than only its PID.

## Why the change works

A bad secret or failed setup stops startup visibly, and a wedged process can no longer pass health just because it still exists.

## Checks

The DAppNode hardening harness covers credential permissions, invalid input, command failure, restart, and health behavior.

### LINKED ISSUE #1658: Bind share proof bundles to canonical signers phases and multiplicity
Audit reference: I14

## What can go wrong

A share-proof bundle could be relabelled to another signer or phase, contain the wrong number of proofs, or replay a conflicting result.

## When it happens

These checks run when C2/C3 share-encryption and C6 share-decryption bundles enter proof verification.

## What changed

Verification now checks signer ownership, committee dimensions, phase layout, proof multiplicity, and per-party replay uniqueness.

## Why the change works

The bundle is accepted only when its cryptographic proofs and its protocol envelope describe the same party, committee, and phase.

## Checks

Relabelling, phase, committee, duplicate, conflict, and C6-count regressions pass.

### LINKED ISSUE #1659: Bind aggregation shares to canonical committee party slots
Audit reference: I15

## What can go wrong

A real committee member could submit an aggregation share while claiming another member's party id.

## When it happens

The risk appears when public-key or plaintext aggregation receives shares and selects the honest subset.

## What changed

The receiver now requires the sender to own the claimed canonical slot and to belong to the selected honest subset.

## Why the change works

A valid network identity can contribute only in its assigned position, preventing one member from occupying multiple logical parties.

## Checks

Slot ownership, subset membership, and canonical-address tests pass.

### LINKED ISSUE #1660: Fail startup closed when EVM event buffers exceed their limit
Audit reference: I16

## What can go wrong

EVM events arriving before startup synchronization completed could grow an unbounded in-memory buffer.

## When it happens

This happens while live subscriptions are active but historical chain backfill has not yet made the node ready.

## What changed

The buffer now has a configurable event limit, and exceeding it fails startup instead of dropping history.

## Why the change works

The node either preserves a complete ordered history within its declared budget or stops for operator recovery; it never silently becomes ready with missing events.

## Checks

Configuration, boundary, overflow, readiness, and actor-stop tests pass.

### LINKED ISSUE #1661: Defer live EVM logs until confirmed canonical backfill
Audit reference: I17

## What can go wrong

Live subscription logs were delivered immediately, bypassing the configured confirmation depth and allowing a later chain reorganization to invalidate work the node had already accepted.

## When it happens

The race is present during and after startup whenever a websocket subscription sees a tip log before canonical backfill reaches the confirmed block.

## What changed

Subscription logs are now wakeups only; confirmed canonical backfill is the sole path that delivers EVM events.

## Why the change works

Every delivered event passes through the same confirmation and ordering rules, regardless of whether a live subscription first noticed it.

## Checks

The confirmed-backfill deferral regression passes.

### LINKED ISSUE #1662: Bound network startup buffering by event count and bytes
Audit reference: I18

## What can go wrong

Network events received before live mode could accumulate forever or be silently evicted, losing protocol history during startup.

## When it happens

This buffer is used while peer traffic is arriving but local recovery and synchronization have not completed.

## What changed

Startup buffering now enforces both event-count and byte ceilings and turns overflow into a readiness failure.

## Why the change works

Memory use is bounded, and the node cannot claim readiness after dropping part of the protocol stream.

## Checks

Count, byte, no-drop, configuration, and readiness tests pass.

### LINKED ISSUE #1663: Drain event snapshot and database state during graceful shutdown
Audit reference: I19

## What can go wrong

Graceful shutdown could close storage while final actor events or snapshots were still queued, so a clean-looking restart could lose the tail of state.

## When it happens

The window begins when the service receives termination and ends when event logs, snapshots, and databases are durably flushed.

## What changed

Shutdown now closes admission, waits for subscriber and actor acknowledgements, drains snapshots, flushes stores, and enforces an internal deadline shorter than the container grace period.

## Why the change works

Each producer is fenced before its destination closes, giving the service manager either a durable stop or a clear timeout failure.

## Checks

Deadline, final-write, acknowledgement, event-log flush, and snapshot-drain tests pass.

### LINKED ISSUE #1664: Preserve large chain finalization delays without integer overflow
Audit reference: I20

## What can go wrong

Converting a large finalization delay from milliseconds to microseconds could overflow and produce the wrong duration.

## When it happens

The conversion happens while chain configuration is loaded, before the EVM synchronization actors start.

## What changed

The duration is now constructed directly from milliseconds.

## Why the change works

No intermediate multiplication is needed, so every valid `u64` millisecond value keeps its meaning.

## Checks

The `u64::MAX` and default-zero regressions pass.

### LINKED ISSUE #1665: Reject internal-only events received through protocol gossip
Audit reference: I21

## What can go wrong

An authenticated peer could send internal infrastructure events through gossip or historical sync even though those events should be created only inside the receiving node.

## When it happens

The boundary is crossed whenever remote gossip or a history page is decoded into an `InterfoldEvent`.

## What changed

Both network paths now use one explicit allowlist of protocol events that are safe to forward.

## Why the change works

Authentication proves who sent a message, not that every internal event type is legal; the allowlist enforces that second rule consistently.

## Checks

Internal-event rejection, allowed-event, and cross-aggregate history tests pass.

### LINKED ISSUE #1666: Make multi-key snapshot writes atomic and acknowledged
Audit reference: I22

## What can go wrong

A snapshot update spanning multiple keys could be only partly written, or reported complete while it was merely queued.

## When it happens

This occurs when actors persist state and its cursor together, particularly during restart-sensitive snapshot updates.

## What changed

The data layer now supports acknowledged atomic and conditional batch writes.

## Why the change works

State and cursor become visible together or not at all, and callers wait until the destination has accepted the complete batch.

## Checks

All-or-nothing, complete-persistence, and partial-key refusal tests pass.

### LINKED ISSUE #1667: Preserve existing node identities during autowallet startup
Audit reference: I23

## What can go wrong

Autowallet startup could overwrite an existing Ethereum or libp2p identity, or leave only one of the pair after an interrupted initialization.

## When it happens

The risk appears on first boot, concurrent boot attempts, or restart with a partially populated data volume.

## What changed

The two identities are initialized atomically; a complete existing pair is preserved and a partial pair is rejected without mutation.

## Why the change works

A node cannot silently change its on-chain or peer identity, and recovery never guesses how to complete ambiguous identity state.

## Checks

Empty initialization, preservation, partial-state, and concurrency tests pass.

### LINKED ISSUE #1668: Isolate snapshot batches by aggregate and await final persistence
Audit reference: I24

## What can go wrong

Snapshot batches from different aggregates collided when they shared a sequence number, and the final flush did not prove that the destination persisted it.

## When it happens

This appears under concurrent E3 activity and at shutdown, when multiple aggregate snapshots can close at similar positions.

## What changed

Batches are now keyed by aggregate plus sequence and their final delivery is acknowledged.

## Why the change works

Unrelated E3s cannot overwrite each other's batch bookkeeping, and shutdown waits for the actual persistence boundary.

## Checks

Cross-aggregate sequence and blocked-destination tests pass.

### LINKED ISSUE #1669: Restore the full logical clock from durable replay history
Audit reference: I25

## What can go wrong

After restart, the hybrid logical clock could resume from an old snapshot and assign timestamps earlier than events already durable in the post-snapshot log.

## When it happens

The gap occurs when a snapshot is stale but later events are replayed before the node returns to live mode.

## What changed

Clock recovery now considers both snapshot state and the maximum complete packed timestamp in replayed history.

## Why the change works

The first new event is ordered after every durable event, including logical-counter ties within the same physical time.

## Checks

Physical/logical time, overflow, restart, and stale-snapshot backlog tests pass.

### LINKED ISSUE #1670: Reject incomplete multi-chain EVM history during startup
Audit reference: I26

## What can go wrong

A multi-chain node could report ready after only one configured chain finished historical synchronization.

## When it happens

This happens during startup backfill when chain RPCs complete or disconnect at different times.

## What changed

Readiness now requires an explicit completion from every configured chain and fails if any chain ends prematurely.

## Why the change works

The node cannot process a partial view of protocol history as though all chains were synchronized.

## Checks

All-chain completion and missing-chain tests pass.

### LINKED ISSUE #1671: Reject incompatible storage schemas before actors write state
Audit reference: I27

## What can go wrong

Actors could interpret unversioned or incompatible bytes as current state and begin writing before anyone checked the on-disk schema.

## When it happens

This is an upgrade and rollback hazard at process startup, before actor construction and replay.

## What changed

A schema preflight now runs first, stamps only genuinely fresh stores, and rejects missing, older, or newer protocol schemas.

## Why the change works

No actor can mutate state until the binary has proved it understands that state, preserving a possible rollback or explicit migration path.

## Checks

Fresh, current, missing, older, newer, and actor-startup tests pass.

### LINKED ISSUE #1672: Bound and exactly consume external bincode payloads
Audit reference: I28

## What can go wrong

An external bincode payload could declare a huge collection or hide extra trailing bytes after an otherwise valid object.

## When it happens

The decoder is used at network and persistence boundaries where payload length or contents may be corrupt or attacker-controlled.

## What changed

Decoding now has byte and allocation limits, uses the expected fixed-integer format, and must consume the input exactly.

## Why the change works

Forged lengths cannot trigger unbounded allocation, and ambiguous payloads with hidden suffixes are rejected.

## Checks

Over-budget, forged-length, trailing-byte, and current-format tests pass.

### LINKED ISSUE #1673: Reject already-expired DHT documents
Audit reference: I29

## What can go wrong

A document whose expiry was already in the past could reach invalid timer arithmetic or be published to the DHT as immediately stale data.

## When it happens

The check is needed when an actor handles a publish request and converts its wall-clock expiry into a runtime deadline.

## What changed

Past expirations are rejected before conversion and before any DHT write.

## Why the change works

Only future deadlines enter scheduling, avoiding underflow-like behavior and preventing useless stale records from propagating.

## Checks

Past/future conversion and no-DHT-write actor tests pass.

### LINKED ISSUE #1674: Bind BFV proofs to exact supported parameter presets
Audit reference: I30

## What can go wrong

Unsupported BFV parameter combinations could select the wrong proof preset, and an invalid polynomial degree could panic during construction.

## When it happens

This boundary is reached when clients build verifiable encryption inputs and when proof code resolves the requested BFV parameters.

## What changed

Only exact supported parameter pairs resolve to a preset, and canonical parameter construction is fallible.

## Why the change works

The encryption parameters and the circuit constants must now come from the same known tuple; malformed inputs return an error instead of diverging or panicking.

## Checks

Noncanonical tuple, invalid degree, supported preset, and variance tests pass.

### LINKED ISSUE #1675: Reject malformed EVM logs before dispatch or checkpointing
Audit reference: I31

## What can go wrong

An EVM log missing its block number or index, failing timestamp lookup, or failing ABI decode could still advance ambiguous chain progress.

## When it happens

The parser sees this during historical backfill and live canonical catch-up, before the log is dispatched and checkpointed.

## What changed

Block number, log index, timestamp, and successful ABI decoding are all required before acknowledgement.

## Why the change works

A cursor advances only for a fully identified, decoded event, so retrying an RPC failure cannot leave a silent hole.

## Checks

Missing metadata, timestamp failure, decode failure, caching, and valid-log tests pass.

### LINKED ISSUE #1676: Reject reverted EVM transaction receipts across all writer paths
Audit reference: I32

## What can go wrong

Code treated the arrival of an EVM receipt as success even when the transaction reverted with status zero.

## When it happens

This affects CLI operations and every actor writer after a submitted transaction is mined.

## What changed

All writer paths now use one successful-receipt validator that checks the status as well as receipt availability.

## Why the change works

A mined revert is surfaced as a failed protocol write instead of allowing local state to assume the contract changed.

## Checks

Receipt-status tests plus utility, EVM, and helper suites pass.

### LINKED ISSUE #1677: Replay aggregates in global timestamp order
Audit reference: I33

## What can go wrong

Replaying each aggregate independently could reorder durable events that originally happened across different E3s.

## When it happens

The problem appears on restart when multiple aggregate logs contain interleaved hybrid logical timestamps.

## What changed

Startup replay now merges aggregates and sorts their events by the packed hybrid logical clock before dispatch.

## Why the change works

Actors observe the same global causal order after restart that the live node recorded before shutdown.

## Checks

The cross-aggregate replay-order regression passes.

### LINKED ISSUE #1678: Bind DHT notification metadata to decoded document payloads
Audit reference: I34

## What can go wrong

A DHT notification could label a valid document as belonging to another E3, chain, document family, or committee party.

## When it happens

The mismatch is possible when network metadata is used to route a decoded DHT payload into protocol actors.

## What changed

The decoded document's family, E3 id, chain id, and party filter are now compared with the notification metadata.

## Why the change works

Routing trusts the signed/decoded payload and metadata only when they agree, so relabelling cannot move data into another protocol context.

## Checks

Cross-E3, cross-chain, and party-filter relabelling tests pass.

### LINKED ISSUE #1679: Backpressure remote network-to-actor ingress
Audit reference: I35

## What can go wrong

Sustained remote traffic could outrun actor mailboxes at the network bridge and grow an uncontrolled queue or lose work.

## When it happens

The pressure point is document, event-translation, and synchronization ingress from libp2p into Actix actors.

## What changed

Those bridges now await destination capacity before accepting more work.

## Why the change works

Backpressure reaches the network task, bounding in-process work by the configured mailboxes instead of buffering indefinitely between subsystems.

## Checks

The complete `e3-net` suite passes; a dedicated long-running saturation drill remains useful follow-up coverage.

### LINKED ISSUE #1680: Defer and coalesce slashing submissions during replay
Audit reference: I36

## What can go wrong

Startup replay could submit slashing transactions before recovery reconciled state, and equivalent evidence could be sent repeatedly in a different order.

## When it happens

This occurs after restart when historical accusation events are replayed through actors that normally perform external effects.

## What changed

External effects stay disabled until reconciliation completes, and submissions are coalesced by the Solidity replay key.

## Why the change works

Replay rebuilds local intent without spending gas; once effects are enabled, each semantic piece of evidence has one idempotency identity.

## Checks

Deferral, reordered-vote, duplicate, retry, and terminal-state tests pass.

### LINKED ISSUE #1681: Constrain program-server callbacks and concurrency
Audit reference: I37

## What can go wrong

The development program server accepted unsafe callback URL forms, followed redirects, and allowed unlimited concurrent compute jobs. A request could therefore redirect unexpectedly, carry credentials in its URL, or consume the test host's workers indefinitely.

## When it happens

The boundary is the development-only `/run_compute` endpoint when a request supplies a callback URL and starts a job. This service is not a production-facing trust boundary and must not be exposed publicly.

## What changed

The linked commit introduced the first hardening pass: callback validation, no redirects, bounded connect/request timeouts, bounded concurrency with HTTP 429, a 10 MiB request cap, and sensitive-log redaction. It also initially added bearer authentication and an exact callback-origin allowlist. The follow-up in #1695 intentionally removed both of those development-only configuration requirements.

The final branch accepts any HTTP(S) callback origin, but rejects URL credentials, fragments, non-HTTP schemes, malformed authority data, and redirect responses. Localhost rewriting remains exact rather than substring based.

## Why the change works

Every job and outbound request has a finite resource budget, and unsafe URL forms or redirect hops are rejected. Because arbitrary HTTP(S) origins are still permitted, deployment isolation—not an application allowlist—is the remaining SSRF boundary.

## Checks

Unsafe URL forms, exact localhost rewriting, redirect policy, timeout configuration, body size, and job-capacity behavior are covered by focused tests.

### LINKED ISSUE #1682: Report supervised child exits and prevent orphaned ciphernodes
Audit reference: I38

## What can go wrong

The supervisor could report an exited child as running, skip graceful termination, orphan a ciphernode, or briefly keep the node's process fence locked across a concurrent child spawn.

## When it happens

These races occur while the local supervisor starts, stops, restarts, or drops its last handle to managed ciphernode processes.

## What changed

Status is refreshed from the OS, shutdown uses TERM then a grace period then KILL, output is drained, partial starts are cleaned up, the last handle kills its child, and the process fence explicitly unlocks on drop.

## Why the change works

Supervisor state now follows the real child lifecycle, every termination path owns cleanup, and fork-time file-descriptor inheritance cannot delay fence reacquisition.

## Checks

Exited-child, TERM-before-KILL, partial-start, orphan-prevention, parallel fence, and repeated reacquisition tests pass.

### LINKED ISSUE #1683: Preserve buffered request-event order and clean terminal state
Audit reference: I39

## What can go wrong

A newly attached request recipient could receive the current event before older buffered events, while buffers for finished E3s could remain in memory.

## When it happens

This appears when an E3-specific actor is created after relevant request events have already arrived, and again when that E3 reaches a terminal state.

## What changed

Buffers are scoped by E3 and recipient, older entries drain before the triggering event, and terminal E3 state is purged.

## Why the change works

Each recipient gets a deterministic per-E3 history in order, and completed requests stop consuming buffer memory.

## Checks

Ordering, isolation, drain, and terminal-cleanup tests pass.

### LINKED ISSUE #1684: Fence startup recovery phases before entering live mode
Audit reference: I40

## What can go wrong

Startup phases could overlap: recovered events might run before effects were safely enabled, or live mode might begin before all recovery history was observed.

## When it happens

The race sits between schema/snapshot recovery, `EffectsEnabled`, event replay, and `SyncEnded` during every restart.

## What changed

The event system now flushes at each phase boundary: effects enablement, recovered history, then live synchronization completion.

## Why the change works

A later phase cannot overtake queued work from the previous phase, so actors see one stable startup sequence.

## Checks

The observable startup-phase ordering regression passes.

### LINKED ISSUE #1685: Count C3 share-encryption proofs by threshold share rows
Audit reference: I41

## What can go wrong

C3 validation counted proof rows using DKG encryption moduli, but honest Shamir shares use the threshold parameter rows. Honest bundles were therefore rejected with the wrong expected proof count.

## When it happens

The mismatch appears when C3 share-encryption proofs are checked after DKG dispatch for a supported committee preset.

## What changed

The dispatched DKG preset is resolved to its threshold counterpart and validation requires the canonical `2 + 2L` C2/C3 proofs.

## Why the change works

The count now comes from the same threshold share shape that generated the bundle, so honest proofs fit exactly and malformed multiplicities still fail.

## Checks

The parameter-domain regression and full actor-level TRBFV flow pass.

### LINKED ISSUE #1686: Use committee-scoped artifacts in the correlated NodeFold E2E test
Audit reference: I42

## What can go wrong

The correlated NodeFold test staged artifacts for the minimum committee but passed a legacy preset-only directory to the prover, causing a misleading test failure unrelated to proof correctness.

## When it happens

This occurs in the NodeFold end-to-end fixture after artifact staging and before proof generation.

## What changed

The prover now receives the same committee-scoped artifact directory that the fixture prepared.

## Why the change works

Staging and consumption refer to one artifact set, so the E2E test exercises the intended minimum-committee circuit.

## Checks

The correlated NodeFold prove-and-verify E2E passes with local `bb` and artifacts.

### LINKED ISSUE #1687: Apply compatible RustSec patches to locked runtime dependencies
Audit reference: I43

## What can go wrong

The lockfile selected vulnerable compatible releases of `bytes`, `crossbeam-epoch`, `quinn-proto`, `rustls-webpki` 0.103, `tar`, and `time` even though patched versions fit the existing manifests.

## When it happens

These packages enter the runtime during a normal locked build; the problem is visible in dependency audit before deployment.

## What changed

The affected lockfile selections were updated within their already-allowed version ranges, without adding overrides or Git dependencies.

## Why the change works

Cargo now selects the upstream patched releases while preserving the public dependency and protocol surface. The separate libp2p/Hickory line is tracked in #1690.

## Checks

Workspace compilation, network tests, and a targeted dependency audit verify the patched selections.

### LINKED ISSUE #1688: Upgrade tracing-subscriber to prevent ANSI log injection
Audit reference: I44

## What can go wrong

The pinned `tracing-subscriber` version allowed user-controlled ANSI escape sequences to alter how retained logs were displayed.

## When it happens

An unsafe sequence can arrive in any remote or structured value that is eventually formatted by the telemetry subscriber.

## What changed

The workspace and support manifest now use `tracing-subscriber` 0.3.20.

## Why the change works

The patched formatter prevents embedded terminal control data from being interpreted as trusted log formatting.

## Checks

Logger redaction/formatting tests and crypto tracing consumers pass.

### LINKED ISSUE #1689: Upgrade Alloy ABI crates to fix typed-data hashing denial of service
Audit reference: I45

## What can go wrong

The pinned Alloy dynamic ABI implementation could consume excessive resources while hashing malicious typed data, creating a denial of service.

## When it happens

The vulnerable path is reached when untrusted structured typed data is parsed and hashed by the ABI stack.

## What changed

Alloy primitives, Solidity types, dynamic ABI, and their coordinated macro crates were upgraded together to 1.4.1.

## Why the change works

Keeping the ABI family on one patched release removes the vulnerable hashing implementation without mixing incompatible generated types.

## Checks

FHE parameter ABI roundtrips, EVM tests, and workspace compilation pass.

### LINKED ISSUE #1692: Synchronize independent Rust consumer lockfiles after Alloy upgrade
Audit reference: I46

## What can go wrong

The CRISP workspace and default template had independent lockfiles that still selected Alloy 1.3 while protocol crates required 1.4.1, so clean CI could not resolve or build them with `--locked`. The root manifest also contained an ignored `workspace.msrv` key, which looked authoritative but Cargo did not use.

## When it happens

This appears in downstream/template CI after the main workspace dependency upgrade, before an E3 integration test can start. The ignored MSRV key can separately mislead maintainers about which compiler version is actually enforced.

## What changed

The independent manifests were aligned with Alloy 1.4.1 and `tracing-subscriber` 0.3.20, only the affected lockfile selections were regenerated, and the ignored root `workspace.msrv` entry was removed.

## Why the change works

Every Rust consumer now resolves the same compatible ABI and logging versions while retaining reproducible locked builds. Removing the unused MSRV declaration prevents a false policy signal; compiler policy must live in tooling Cargo actually reads.

## Checks

The default support-script locked build and the CRISP wasm-target locked check pass.

### LINKED ISSUE #1693: Allow schema preflight after atomic bootstrap identity creation
Audit reference: Follow-up

## What can go wrong

A brand-new node creates its Ethereum and libp2p identities before schema preflight. The preflight mistook those two bootstrap keys for legacy protocol state and stopped every fresh node with a missing-schema error.

## When it happens

This happens on the first startup after atomic autowallet initialization, before actors or history replay begin.

## What changed

Preflight now treats either an empty store or exactly one complete identity pair as fresh and stamps the current schema. Partial identities, extra keys, snapshots, or event-log records still fail closed, and the same rule is used by the storage validator.

## Why the change works

Bootstrap identity is recognized narrowly without weakening upgrade safety: anything beyond the exact known pair still requires an existing schema marker.

## Checks

Focused in-memory, Sled, validator, and actor-level startup tests pass.

### LINKED ISSUE #1695: Remove bearer authentication from the development program server
Audit reference: Follow-up

## What can go wrong

Bearer-token and exact callback-origin configuration coupled every local launcher and caller to credentials and URLs without creating a meaningful production security boundary. Fresh CI and developer environments failed when those values were absent.

## When it happens

It occurs when the development-only program server starts or a local CRISP/template caller submits a compute request without matching configuration.

## What changed

The token field, builder requirement, authorization check, environment variables, caller headers, and exact callback-origin allowlist were removed. The service still rejects callback credentials, fragments, non-HTTP schemes, malformed authorities, and redirects; it keeps bounded connect/request timeouts, the 10 MiB body cap, bounded jobs, exact localhost rewriting, and log redaction.

## Why the change works

Local/test consumers work without pretend production authentication or brittle origin coupling, while the URL-form and resource controls still reduce accidental misuse. Because any HTTP(S) callback origin is allowed, the service remains explicitly development/test-only and must be isolated from untrusted clients and sensitive networks.

## Checks

The server starts without a token, callback validation exercises the final URL policy, and both CRISP and template callers submit token-free requests in focused tests.

### LINKED ISSUE #1696: Restore finalized committee slots before C6 verification after restart
Audit reference: Recovery follow-up

## What can go wrong

After an aggregator restarted, its C6 verifier had no in-memory map from committee members to canonical party slots. The finalization event could already be inside the snapshot, so replay would not send it again and honest decryption shares would be rejected. Keeping every finalized committee forever would fix restart lookup at the cost of unbounded durable growth.

## When it happens

The failure occurs after committee finalization and public-key publication when the active aggregator restarts before C6 share verification completes. Retention cleanup occurs later when the E3 reaches a terminal success or failure stage.

## What changed

Before replay, the builder restores finalized committee slots from the durable sortition repository. Live and replayed lifecycle events keep the verifier map current. Terminal lifecycle handling now removes the completed committee from the verifier and durable recovery state, with bounded retention logic pruning older finalized committees as well.

## Why the change works

The verifier has the same canonical party ownership after restart even when `CommitteeFinalized` is outside the replay window, so honest C6 bundles follow the normal consistency checks. Terminal deletion and bounded retention prevent the recovery index from growing forever.

## Checks

Focused restart tests omit the finalization event deliberately and prove restored honest C6 bundles advance. Separate retention tests cover terminal removal and pruning behavior.

### LINKED ISSUE #1697: Restore C0 verification context before replay after restart
Audit reference: Recovery follow-up

## What can go wrong

The C0 proof verifier kept committee slots and E3 BFV presets only in memory. After a snapshot restart, the lifecycle events that populated them could be outside replay, causing honest in-flight C0 proofs to be rejected.

## When it happens

This occurs after `CommitteeFinalized` and `CiphernodeSelected`, when a node restarts before all C0 key-generation proofs are verified.

## What changed

Finalized slots and the exact E3 parameter preset are restored from durable repositories before EventStore replay; malformed identities and unsupported parameters fail closed, and terminal cleanup remains active.

## Why the change works

C0 verification receives the same request and committee context it had before restart without depending on already-snapshotted lifecycle events being replayed twice.

## Checks

Focused restart tests intentionally omit those lifecycle events and prove an honest canonical C0 submission reaches normal verification; workspace tests and strict checks pass.

### LINKED ISSUE #1699: Remove the redundant CiphernodeHandle constructor
## What was difficult to maintain

`CiphernodeHandle` had a ten-argument constructor whose only job was to assign each argument to the matching struct field. Several arguments had the same or similar types, so call sites depended on positional ordering and were harder to review safely.

## When it matters

This comes up whenever the ciphernode builder gains, removes, or reorders a runtime handle such as the event bus, network interface, event store, or aggregate IDs. A positional constructor makes an otherwise small builder change touch two definitions and makes swapped values less obvious during review.

## What changed

The redundant constructor was removed and the builder now constructs `CiphernodeHandle` with named fields at the single call site.

## Why this is simpler

Each value is visibly paired with its destination field, adding a field requires changing only the struct literal, and 26 lines of forwarding code are gone without changing runtime behavior.

## Checks

The locked workspace build, strict Clippy run, formatting checks, and pre-push lint suite pass.

### LINKED ISSUE #1700: Consolidate the ciphernode architecture and audit ledger
## What was difficult to understand

The ciphernode architecture, crate responsibilities, recovery flow, and production-readiness audit decisions were spread across older scenarios and several partial flow notes. That made it difficult to tell which document was current or trace a protocol event through the crates that handle it.

## When it matters

This affects code review, incident investigation, onboarding, and future refactoring. It is especially costly when reviewing recovery or lifecycle changes because the relevant actor, durable state, and terminal cleanup behavior cross multiple crates.

## What changed

A consolidated crate architecture and audit ledger was added under `agent/CRATES_ARCHITECTURE.md`. The agent rules and flow-trace index now point to it as the source of truth, and the DKG/computation plus deactivation/completion traces were updated to match the final audited implementation.

## Why this helps

Reviewers can start from one maintained map, follow a feature to its owning crate and flow trace, and distinguish implemented safeguards from remaining production work without reconstructing the design from historical scenario files.

## Checks

The documented crate and flow references were checked against the final PR tree, and the documentation commit passes patch-integrity and pre-push repository checks.

hmzakhalid:
## Summary

This draft PR carries the production-readiness and Rust architecture pass for the ciphernode and its protocol-critical dependencies.

- fixes replay crash loops and propagates backpressure through the startup event pipeline;
- hardens durable storage, schema preflight, snapshot atomicity, logical-clock recovery, shutdown draining, and supervisor cleanup;
- bounds and validates network ingress, historical synchronization, startup buffers, DHT metadata, and external serialization;
- strengthens EVM confirmation, parsing, nonce, receipt, replay, and slashing-effect behavior;
- binds proof/share aggregation to canonical E3, signer, committee, party, phase, parameter, and multiplicity invariants;
- keeps the development/test program server token-free, removes callback-origin setup, and retains bounded webhook delivery, payload logging, and job concurrency;
- updates compatible vulnerable dependencies without adding release-policy or supply-chain test machinery;
- adds the final 45-package runtime and protocol architecture map under agent/.

No Solidity public ABI, deployed storage layout, Noir circuit source, circuit semantics, wire format, production deployment, key, secret, fund, release, or publication action is included.

## Validation

- pnpm rust:test: PASS, 1,108 listed tests; real 1,021.30s; no failures.
- e3-tests actor integration: 3 passed, 2 ignored; TRBFV completed in 354.79s.
- correlated NodeFold proof/verify E2E: 1 passed in 82.01s.
- cargo check --workspace --all-targets --locked: PASS in 1m41s.
- cargo clippy --workspace --all-targets --locked -- -D warnings: PASS in 1m03s.
- cargo test --workspace --doc --locked: PASS.
- cargo fmt --all -- --check and git diff --check: PASS.
- pre-push repository lint, pnpm/license, and committee consistency hooks: PASS; existing Solidity lint emitted 3 warnings and nargo was unavailable, so the hook skipped Noir lint.
- cargo audit --no-fetch: reduced from 18 to 5 vulnerabilities; remains non-zero and is tracked by #1690.
- circuit diff versus main: empty.
- default-template locked support build and CRISP wasm-target locked check: PASS.
- token-free and callback-origin-free program-server regressions: PASS (8 Rust server tests); focused CRISP and template token-free regressions previously passed.
- C0/C6 restart recovery regressions (restored_): PASS (2 tests; 128 filtered).

Docker-backed disk-full/partition fault injection was not run because the Docker daemon is unavailable. The standalone crates/support-scripts/dev manifest remains blocked by its pre-existing workspace-membership error.

## Fixed issues

- Closes #1645 — Bind C0 proofs to the canonical E3 request and committee slot
- Closes #1646 — Use the honest-party quorum for accusation voting
- Closes #1647 — Fail closed on corrupt event logs and reconcile indexes in bounded pages
- Closes #1648 — Keep network broadcast consumers alive after receiver lag
- Closes #1649 — Make startup replay end-to-end backpressured
- Closes #1650 — Validate canonical H-of-N shapes before DKG aggregation
- Closes #1651 — Match the SDK finalized-committee event to the contract ABI
- Closes #1652 — Bound ciphernode startup with a configurable readiness deadline
- Closes #1653 — Serialize pending nonce allocation for each chain signer
- Closes #1654 — Replace probabilistic event deduplication and reject timestamp conflicts
- Closes #1655 — Redact and bound sensitive structured tracing fields
- Closes #1656 — Bound historical peer synchronization and require cursor progress
- Closes #1657 — Fail DAppNode startup closed on invalid credentials and false health
- Closes #1658 — Bind share proof bundles to canonical signers phases and multiplicity
- Closes #1659 — Bind aggregation shares to canonical committee party slots
- Closes #1660 — Fail startup closed when EVM event buffers exceed their limit
- Closes #1661 — Defer live EVM logs until confirmed canonical backfill
- Closes #1662 — Bound network startup buffering by event count and bytes
- Closes #1663 — Drain event snapshot and database state during graceful shutdown
- Closes #1664 — Preserve large chain finalization delays without integer overflow
- Closes #1665 — Reject internal-only events received through protocol gossip
- Closes #1666 — Make multi-key snapshot writes atomic and acknowledged
- Closes #1667 — Preserve existing node identities during autowallet startup
- Closes #1668 — Isolate snapshot batches by aggregate and await final persistence
- Closes #1669 — Restore the full logical clock from durable replay history
- Closes #1670 — Reject incomplete multi-chain EVM history during startup
- Closes #1671 — Reject incompatible storage schemas before actors write state
- Closes #1672 — Bound and exactly consume external bincode payloads
- Closes #1673 — Reject already-expired DHT documents
- Closes #1674 — Bind BFV proofs to exact supported parameter presets
- Closes #1675 — Reject malformed EVM logs before dispatch or checkpointing
- Closes #1676 — Reject reverted EVM transaction receipts across all writer paths
- Closes #1677 — Replay aggregates in global timestamp order
- Closes #1678 — Bind DHT notification metadata to decoded document payloads
- Closes #1679 — Backpressure remote network-to-actor ingress
- Closes #1680 — Defer and coalesce slashing submissions during replay
- Closes #1682 — Report supervised child exits and prevent orphaned ciphernodes
- Closes #1683 — Preserve buffered request-event order and clean terminal state
- Closes #1684 — Fence startup recovery phases before entering live mode
- Closes #1685 — Count C3 share-encryption proofs by threshold share rows
- Closes #1686 — Use committee-scoped artifacts in the correlated NodeFold E2E test
- Closes #1687 — Apply compatible RustSec patches to locked runtime dependencies
- Closes #1688 — Upgrade tracing-subscriber to prevent ANSI log injection
- Closes #1689 — Upgrade Alloy ABI crates to fix typed-data hashing denial of service

- Closes #1692 — Synchronize independent Rust consumer lockfiles after Alloy upgrade

## Follow-up risk

- #1690 tracks the five remaining libp2p-transitive RustSec advisories and the required tested libp2p/TLS/DNS migration.
- Additional protocol and operational residuals that need design approval or external fault infrastructure remain tracked in linked follow-up issues.

### CI follow-up fixes

- Closed #1681 as not planned — callback-origin allowlisting was removed for the development/test server.
- Closes #1693 — Allow schema preflight after atomic bootstrap identity creation
- #1694 — Wire program-server authentication through local development consumers; superseded and reverted by #1695
- Closes #1695 — Remove bearer authentication from the development program server
- Closes #1696 — Restore finalized committee slots before C6 verification after restart
- Closes #1697 — Restore C0 verification context before replay after restart
- Closes #1699 — Remove the redundant CiphernodeHandle constructor
- Closes #1700 — Consolidate the ciphernode architecture and audit ledger


crates/zk-prover/src/actors/proof_verification.rs
```diff
@@ -181,9 +234,34 @@ impl Handler<InterfoldEvent> for ProofVerificationActor {
...
+                            e3_id = %data.e3_id,
+                            "Finalized committee contains an invalid address: {err}; C0 keys will be rejected"
+                        );
+                    }
+                }
+            }
             InterfoldEventData::EncryptionKeyReceived(data) => {
                 self.notify_sync(ctx, TypedEvent::new(data, ec))
             }
+            InterfoldEventData::E3RequestComplete(data) => {
```
ctrlc03: do we need similar for E3Failed ?

ctrlc03 (approved):

PR #1706, opened 2026-07-15
chore: update template [skip-line-limit]

### LINKED ISSUE #748: Template app should get back to the same point after refresh
Refreshing the page resets the app state and forces the user to start over. This disrupts the flow when testing or following the wizard.

The app should persist relevant state so that after a browser refresh, the user can continue from the same point without repeating previous steps.

### LINKED ISSUE #749: Template app should show loaders when UI is not usable
Right now the UI can be clicked even while data are still loading, which leads to errors or confusion. The app should display loaders or disable interactions during async operations (e.g. initialization, contract transactions, data fetch) to make the flow clearer.

Acceptance Criteria:
	•	Show loading states for async actions
	•	Disable or block UI interactions until the action completes
	•	Provide clear feedback (spinner, overlay, etc.)

### LINKED ISSUE #750: Minimize server code in the template
The template server currently contains logic that could run fully on the client. This increases complexity and makes the server harder to maintain. The goal is to migrate as much logic as possible to the client/SDK and keep the server minimal—only for cases where a backend is strictly required.

### LINKED ISSUE #1705: Update template app design to match the rest of the apps

ctrlc03:
fix #749 fix #748 fix #1705 fix #750


0xjei (approved):
utack

PR #1733, opened 2026-07-19
feat(cli): add --linux-network-host-mode for program start

### LINKED ISSUE #548: Add flag: `enclave program start --linux-network-host-mode`
Allow users to run enclave program start using --network=host mode in docker as opposed to rewriting callback urls and adding host.local to the host file. 

This avoids issues with linux users having to tweak their firewall rules unnecessarily.

crazywriter1:
## Summary
- Add `--linux-network-host-mode` to `interfold program start` for Linux Docker host networking
- Skip `localhost` → `host.local` rewrite when the flag is set (`INTERFOLD_SKIP_LOCALHOST_REWRITE`)
- Document the flag in `crates/support/README.md`

Closes #548

## Test plan
- [x] `CARGO_BUILD_JOBS=1 cargo build -p e3-cli`
- [x] `./target/debug/interfold program start --help | grep -A2 linux-network-host-mode`
- [x] `ctl/start` / `ctl/container` dry-runs with stub docker

## Summary by CodeRabbit

* **New Features**
  * Added optional Linux host-networking mode when starting the program server.
  * Preserves callback URLs correctly when host networking is enabled.
  * Development mode warns that host networking has no effect.
  * The option is rejected on non-Linux systems; standard networking remains unchanged.

* **Bug Fixes**
  * Disabled automatic redirects for webhook delivery.
  * Improved localhost callback handling for host-networked environments.

* **Documentation**
  * Added Linux-specific setup instructions, platform limitations, and callback behavior.


Toby1009 (approved):
LGTM
Please update this branch with the latest main before merging.

crazywriter1:
I've updated the branch with the latest main. Please let me know if you'd like me to address the remaining CodeRabbit suggestions as well. @Toby1009

Toby1009:
> I've updated the branch with the latest main. Please let me know if you'd like me to address the remaining CodeRabbit suggestions as well. @Toby1009

I’d lean toward fixing these, but maybe we should also ask @ctrlc03 to take a look.

crates/support-scripts/src/program_dev.rs
```diff
@@ -26,7 +26,12 @@ impl ProgramSupportApi for ProgramSupportDev {
         run_bash_script(&cwd, &script, &[]).await?;
         Ok(())
     }
-    async fn start(&self) -> Result<()> {
+    async fn start(&self, linux_network_host_mode: bool) -> Result<()> {
+        if linux_network_host_mode {
+            eprintln!(
```
ctrlc03: use tracing::warn!

crates/support-scripts/ctl/container
```diff
@@ -40,22 +40,44 @@ else
...
-    # in the docker container we can then replace localhost and 127.0.0.1 
-    # from the input callback url so calls redirect to gateway.
-    # This should in theory be crossplatform
-    # However on linux the user must allow incoming connections from Docker's bridge network 172.17.0.0/16 through their firewall.
+    # Default (cross-platform): `--add-host=host.local:host-gateway` plus rewriting
+    # localhost/127.0.0.1 in callback URLs so the container can reach the host.
+    # On Linux that requires allowing Docker's bridge network (172.17.0.0/16) through
+    # the firewall. `--linux-network-host-mode` uses Docker `--network=host` instead
+    # so localhost callbacks work without rewrite or firewall changes.
+    # Note: `--network=host` is not useful on macOS Docker Desktop.
```
ctrlc03: let's add some checks if the system is mac 

if [[ "$LINUX_NETWORK_HOST_MODE" == "1" && "$(uname -s)" != "Linux" ]]; then
    echo "Error: --linux-network-host-mode needs Linux. Detected $(uname -s)." >&2
    echo "Docker Desktop does not enable --network=host by default, and the container" >&2
    echo "port would not be published. Start again without the flag." >&2
    exit 1
fi

crazywriter1: Addressed in 68e713c: added the non-Linux guard right after flag parsing in ctl/container.

crazywriter1:
@ctrlc03 Thanks for the review sir. Addressed your two suggestions in the latest push (`tracing::warn!` + non-Linux guard).

CodeRabbit also suggested container mode mismatch handling and callback URL validation. Should I address those in this PR as well?

ctrlc03:
> @ctrlc03 Thanks for the review sir. Addressed your two suggestions in the latest push (`tracing::warn!` + non-Linux guard).
> 
> CodeRabbit also suggested container mode mismatch handling and callback URL validation. Should I address those in this PR as well?

yes please can you address coderabbit comments too?

crazywriter1:
Addressed the remaining CodeRabbit feedback:
- recreate the support container when network mode changes
- validate callback URLs before webhook dispatch @ctrlc03

crazywriter1:
Applied redirect Policy::none() to match e3-program-server. The DNS resolution/pinning suggestions go beyond what program server does today. Happy to discuss separately if you want stricter SSRF policy across both servers.

ctrlc03:
> Applied redirect Policy::none() to match e3-program-server. The DNS resolution/pinning suggestions go beyond what program server does today. Happy to discuss separately if you want stricter SSRF policy across both servers.

please make sure you fix all coderabbit issues. thanks

crazywriter1:
@ctrlc03 All CodeRabbit comments addressed in 0410a5d. Container recreation, callback validation, redirect blocking, DNS pinning, IPv4-mapped IPv6 handling, and test fixes. Ready for review.

Toby1009:
Maybe I think the callback security changes have grown beyond what #548 needed, and
it may be better to split them out.

#548 asked for an optional Linux host-networking flag. The PR now adds 412 lines,
including 280 additions to e3-support-app, most of which implement callback
validation, DNS resolution, pinning, and related tests.

Three concerns:

1. `agent/CRATES_ARCHITECTURE.md:596` documents the equivalent e3-program-server
   endpoint as accepting caller-supplied HTTP(S) callback origins, and
   crates/program-server/src/lib.rs:182 implements exactly that. This PR
   introduces a substantially different policy in e3-support-app, without
   documenting why the two servers should have different callback trust
   boundaries.

2. The new policy does not address the main callback risk introduced by host
   networking. Host mode makes host loopback-only services reachable. The
   host-mode branch accepts any loopback callback URL, including arbitrary ports
   and paths, so it does not distinguish the intended callback service from other
   host-local services. Meanwhile, the default bridge path now rejects RFC1918
   destinations (LAN, VPN, Compose service names), although this PR does not
   change that path's network reachability.

3. The IP policy is difficult to maintain as a denylist. I re-ran CodeRabbit
   against the current head and it asks for five more CIDR ranges, so after four
   rounds it still classifies several non-global special-use ranges as safe.
   Also, crates/support is a separate workspace excluded from the root workspace
   (Cargo.toml:54), and its new callback tests are not currently run by CI.

I suggest keeping the flag, plumbing, non-Linux guard, container recreation, and
the contained Policy::none() change in this PR. DNS resolution, private-address
filtering, and pinning could move to a separate security issue that covers both
servers and starts by defining caller authentication and configu
origins.

Two bugs worth fixing either way:
- the existing-container path always executes `exit 0` after `do
  failures inside the container are reported as success (the script has no
  `set -e`)
- the mode-switch path falls through from `docker stop` to `docker run --name`;
  `--rm` removal is async, so switching modes can hit a name con

Happy to be corrected if I have misunderstood the intended trust

crazywriter1:
@Toby1009 Thanks for the detailed review. The callback security work grew because @ctrlc03 asked me to address the remaining CodeRabbit comments; I wasn't aiming to expand #548 beyond the flag.

I agree it's gotten large and the policy inconsistency with e3-program-server is worth discussing separately. Happy to split the security changes out and keep this PR to flag + plumbing + container recreation + Policy::none(), and fix the two container script bugs either way. Leaving the call to you and @ctrlc03. Just say which direction and I'll update the branch.

crazywriter1:
@Toby1009 @ctrlc03 I’m stuck on the merge state.

I merged main to fix the earlier conflict, then main got force-pushed and GitHub shows ~150 conflicts again — mostly in files I never touched.

How should I proceed: recreate the branch from current main with only the host-mode changes, or rebase?

Toby1009:
> @Toby1009 @ctrlc03 I’m stuck on the merge state.
> 
> I merged main to fix the earlier conflict, then main got force-pushed and GitHub shows ~150 conflicts again — mostly in files I never touched.
> 
> How should I proceed: recreate the branch from current main with only the host-mode changes, or rebase?

@crazywriter1  I still lean towards trimming this back to what #548
asked for and splitting the callback hardening out.

The tests for the new validation don't run anywhere as far as I can tell.
`crates/support` isn't in the root workspace (`Cargo.toml:54`), so the `cargo
test` step (`.github/workflows/ci.yml:174`) doesn't reach it — and that step is
`--lib` only, while `e3-support-app` is a binary, so it wouldn't pick them up
even if the crate were in the workspace. The only other thing CI does with the
crate is build the Docker image (`ci.yml:315`), and that Dockerfile doesn't
compile anything. So the 9 tests in `crates/support/app/src/main.rs:406-478`
look like dead weight right now.

The IP denylist still isn't settled after three CodeRabbit rounds, and these are
usually hard to converge — an allowlist of callback origins would at least
terminate. It also leaves the two servers with different rules: there's a test
at `crates/program-server/src/lib.rs:449` asserting
`https://metadata.internal:8443/latest` is a valid callback, which is exactly
what support-app would now reject. Could be deliberate — support-app ships in a
container and program-server doesn't — but if so it'd be worth writing down.

Here's what I'd keep, if trimming makes sense:

- `aa19fbf2` and `59dc260c`
- `2601b0a5` (`Policy::none()`) — 5 lines, and it matches what program-server
  already does
- the `ctl/container` + `Cargo.lock` half of `99a36647`, dropping the `main.rs`
  half

`13e4cdfd` and `46929708` would come out. Happy to write up the follow-up issue
for them.

Two things in `ctl/container`, worth fixing whichever way this goes:

- `exit 0` at line 82 — there's no `set -e` in the script, so it throws away
  what `docker exec` returned and failures inside the container come back as
  success. On main the exec was the last command so its status propagated;
  looks like `99a36647` introduced this. `exit $?` should do it.
- the mode switch at line 78 stops the container and falls through to `docker
  run --rm --name` at 103-104. `--rm` cleans up asynchronously, so the name
  might not be free yet and that run can hit a conflict.

crazywriter1:
Hey @Toby1009, trimmed it down as you suggested.

Kept:
- `aa19fbf2` and `59dc260c` (host-mode flag + plumbing)
- `Policy::none()` from `2601b0a5`
- The `INTERFOLD_SKIP_LOCALHOST_REWRITE` conditional in `ctl/container`

Dropped `13e4cdfd` and `46929708` entirely — URL validation, DNS pinning, IP filtering, and the callback tests are all gone from this PR.

Also fixed the two `ctl/container` bugs you flagged:
- `exit $?` instead of `exit 0` so `docker exec` failures propagate
- Bounded wait (5 s) after `docker stop` before `docker run --name` to avoid the `--rm` async name conflict

crazywriter1:
@coderabbitai resume

PR #1748, opened 2026-07-28
docs(crisp): correct and complete the local setup instructions

Toby1009:
# docs(crisp): correct and complete the local setup instructions

Documentation only — no behaviour changes.

Every item below was hit or verified while bringing the CRISP example up from a clean checkout on
macOS 15 (arm64).

## Missing prerequisites

- **`git submodule update --init --recursive`** was not mentioned anywhere. Without it,
  `pnpm dev:setup` fails compiling contracts with
  `HHE902 ... lib/risc0-ethereum/contracts/src/groth16/RiscZeroGroth16Verifier.sol doesn't exist`.
  CI never hits this because its checkout uses `submodules: recursive`. Verified on a fresh clone:
  `git submodule status` shows both entries uninitialised and the `.sol` file absent.
- **`yq`** was not listed, but `scripts/dev_cipher.sh` reads every node address with it. Without it
  the script aborts with `yq: command not found` under `set -euo pipefail`. `crisp_e2e` has an
  explicit `Install yq` step, so CI already knows it is required.
- **Pinned `nargo` / `bb` versions.** The README linked to the tools without versions. Added the
  versions CI pins (`NOIR_TOOLCHAIN` = `v1.0.0-beta.16`) and a pointer to
  `crates/zk-prover/versions.json`, which is the source of truth for the bb version and its
  per-platform checksums.
- **Node version.** CI uses 22.x; odd-numbered releases make Hardhat print an unsupported-version
  warning on every invocation. (It is a warning, not a hard failure — the stack does run on 25.x.)

## Corrections

- **RiscZero is not a prerequisite for local development.** `scripts/dev_program.sh` runs
  `interfold program start --dev true` unconditionally, so `pnpm dev:up` never uses a proving
  backend regardless of `program.dev` in `interfold.config.yaml`. The real-proving path runs inside
  a container image that already ships RiscZero, so what is actually needed locally for that flow is
  Docker. I brought the full stack up and completed a DKG round without ever installing
  `rzup`/`r0vm`.
- **`dev:setup` does not apply `crisp.dev.env` to `server/.env`.** The comment claimed
  `applies crisp.dev.env → server/.env`, but `apply_crisp_dev_config_to_server_env` only copies
  `server/.env.example` when the file is absent and writes none of the profile values.
- **The standalone `dev_cipher.sh` command was missing its argument.** The README showed
  `./scripts/dev_cipher.sh`, which fails immediately with `$1: unbound variable` because the script
  runs under `set -u` and reads `READYFILE=$1`.

`prettier --check` passes.


ctrlc03 (approved):
Thanks

PR #1755, opened 2026-07-28
fix(docs): repair broken logo path in the root README

Toby1009:
# fix(docs): repair broken logo path in the root README

The header image at the top of the root `README.md` points at `./interfold-meta.jpg`, which does not
exist at the repository root, so the logo renders as a broken image on the repository landing page.

The file only exists at `docs/public/interfold-meta.jpg`. It was added there for the docs site during
the Interfold rename (#1402), but the README was pointed at a root-level copy that was never added.
The root has never contained a `.jpg` — `git log --diff-filter=D -- '*.jpg'` returns nothing, so this
is not a file that went missing later.

## Verification

Against what GitHub actually requests when rendering the README on `main`:

| URL | Status |
| --- | --- |
| `raw/main/interfold-meta.jpg` (what the README points at) | **404** |
| `raw/main/docs/public/interfold-meta.jpg` (where the file is) | 200 |

## Why repoint rather than copy

Repointing the `src` avoids committing a second copy of the image. Adding a root-level duplicate
would work equally well for rendering, but leaves two files to keep in sync.

`prettier --check README.md` passes.

## Unrelated observation

`enclave.png` (1.3 MB) is still tracked at the repository root and is referenced by no markdown file
(`git grep enclave.png -- '*.md'` returns nothing) — apparently left over from the same rename. Not
touched here, but it looks like it could be removed.

## Summary by CodeRabbit

* **Documentation**
  * Updated the README hero image reference so the image displays correctly from its new location.


Toby1009:
One open question before this is merged — I kept the image the README was already pointing at, but
I am not sure it is the right asset, so I would rather ask than decide it for you.

`interfold-meta.jpg` is 1030 KB at 1200×628. The dimensions and the `-meta` name suggest it is the
Open Graph / social preview card rather than a logo, so pointing the README header at it means the
landing page and the social card share one asset.

For context, here is what is currently in the repo:

| Asset | Size | Dimensions | Apparent role |
| --- | --- | --- | --- |
| `enclave.png` (root) | 1306 KB | 1456×432 | previous README header, before the rename |
| `docs/public/interfold-meta.jpg` | 1030 KB | 1200×628 | what the README points at today |
| `docs/public/interfold-logo.png` | 5 KB | 500×72 | wordmark |
| `docs/public/enclave-meta.jpg` | 155 KB | 1200×628 | previous branding's social card |

Two things stand out:

- This is **not** a size regression. The header it replaces (`enclave.png`) was 1306 KB, so 1030 KB
  is already an improvement over what was rendering before the rename.
- But at identical dimensions the previous social card was 155 KB, so ~1 MB looks more like a
  missed compression pass than a deliberate choice.

So: keep this image, or use something else? Options as I see them:

1. **Keep it as-is** — matches what the README pointed at, and is smaller than the pre-rename header.
2. **Keep it but recompress** — `enclave-meta.jpg` shows 1200×628 fits comfortably in ~155 KB.
3. **Use `interfold-logo.png` instead** — 5 KB, and it is an actual logo rather than a social card.

Happy to update this PR either way. This is a branding call, so I did not want to make it unilaterally.

ctrlc03 (approved):

PR #1782, opened 2026-08-10
docs: fix BFV parameter decoding in Hello World tutorial

CryptAm:
## Summary

Fix BFV parameter decoding in the Hello World tutorial.

## Changes
- Added `.unwrap()` to both `decode_bfv_params_arc()` calls in `docs/pages/hello-world-tutorial.mdx`. 

## Rationale

`decode_bfv_params_arc()` returns `Result<Arc<BfvParameters>, EncodingError>`, so its result must be handled before being passed to `Ciphertext::zero()` and `Ciphertext::from_bytes()`.

The current implementation in `crates/fhe-params/src/encoding.rs` and its tests explicitly handle the result, while the generated `templates/default/program/src/lib.rs` already uses `.unwrap()`.

This change aligns both examples with the existing implementation.

## Summary by CodeRabbit

* **Documentation**
  * Updated Rust examples in the Hello World tutorial to use decoded FHE parameters correctly in ciphertext operations.


ctrlc03 (approved):
Thanks

PR #1821, opened 2026-08-17
fix: crisp security issues [skip-line-limit]

ctrlc03:
Huge thanks to @Toby1009 for reporting most of these issues. 

## Summary by CodeRabbit

* **New Features**
  * Added ciphertext commitments and slot metadata to compute inputs.
  * Added append-only input tracking with latest-valid-entry selection.
  * Added provenance manifests and deployment verification tooling.
  * Added optional ciphertext-verifier configuration for protocol deployments.
  * Added commitment generation to the CRISP SDK and WASM interfaces.

* **Bug Fixes**
  * Invalid, malformed, or mismatched inputs are rejected with clear errors.
  * Enforced two-component ciphertexts and improved fallback handling for invalid entries.

* **Documentation**
  * Added guidance for verifying deployments, guest provenance, input-root consistency, and secure compute programs.

## Summary by CodeRabbit

- **New Features**
  - Added commitment-bound input metadata, including slots and parent links.
  - CRISP now supports first votes, re-votes, masking, and append-only histories.
  - Added deployment verification tools and compute-provider provenance manifests.
  - Updated SDK and APIs to expose previous ciphertexts with their tree positions.

- **Bug Fixes**
  - Improved validation of malformed ciphertexts, commitments, poisoned entries, and incomplete snapshots.
  - Enforced zero plaintext for masking operations and stricter ballot boundaries.
  - Added checks for verifier, guest image, and toolchain consistency.

- **Documentation**
  - Added deployment verification, provenance, secure-compute, and CRISP workflow guidance.


crates/support/methods/guest/Cargo.toml
```diff
@@ -13,6 +13,9 @@ path = "src/bin/program.rs"
 
 [dependencies]
 risc0-zkvm = { version = "=3.0.3", default-features = false, features = ['std'] }
+# This revision decides which compute-provider code runs inside the zkVM, so it decides the image
+# ID. It must match the pin in crates/support/Cargo.toml. See that file for why the dependency is
+# pinned by revision instead of by path.
 e3-compute-provider = { git = "https://github.com/theinterfold/interfold", rev = "c2097da61b4d07c4ce83840393ff4e9f171eefb4" }
```
Toby1009: This workspace's lockfile is one line behind its manifests, so the reproducible build stops before it
starts:

```
$ cargo metadata --locked --offline --manifest-path crates/support/methods/guest/Cargo.toml
error: the lock file ... needs to be updated but --locked was passed to prevent this
```

Letting cargo repair it gives a one-line diff: `alloy-primitives` under `e3-support-types`.
`crates/support/types` gained that dependency in 02ebe458, which regenerated
`crates/support/Cargo.lock` but not this one.

It stayed invisible because `risc0-build` passes `--locked` only on the Docker path. A local build
repairs the lockfile and carries on, so the failure surfaces exactly where
`verifying-the-compute-provider.mdx` sends a reader.

scripts/check-image-id.sh
```diff
@@ -0,0 +1,202 @@
...
+if [ "$pin_count" -ne 1 ]; then
+  fail "the guest and host workspaces pin different Interfold revisions:
+$(printf '%s\n' "$pins" | sed 's/^/    /')
+  Every Interfold dependency in crates/support must name one revision."
+fi
+PINNED_REV="$pins"
+
+# The pin must be a commit this repository knows, so a reviewer can diff it against the tree.
+if git cat-file -e "${PINNED_REV}^{commit}" 2>/dev/null; then
+  if ! git merge-base --is-ancestor "$PINNED_REV" HEAD 2>/dev/null; then
```
Toby1009: This asks whether the pinned revision is an ancestor of HEAD, which catches a revision the tree does
not know. Staleness is a different question, and it currently passes:

```
$ git merge-base --is-ancestor c2097da6 origin/main   # exit 0, no warning
$ git rev-list --count c2097da6..origin/main          # 47
$ git log --oneline c2097da6..origin/main -- crates/compute-provider | wc -l   # 4
```

The comment added to `crates/support/Cargo.toml` says this revision "predates fixes that have since
landed on main", so the knowledge is already here. It lives in a comment rather than in the check.
The same command answers the other question with different arguments:

```bash
last=$(git log -1 --format=%H -- crates/compute-provider crates/support/types)
git merge-base --is-ancestor "$last" "$PINNED_REV" || warn "pin predates the last change to those paths"
```

On this branch that returns 1.

crates/support/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG RUST_VERSION=1.91.1
 ARG RISC0_VERSION=3.0.3
-ARG RISC0_TOOLCHAIN=1.88.0
+ARG RISC0_TOOLCHAIN=1.91.1
```
Toby1009, line 3: Not about the value. This `ARG` sits before the first `FROM`, and the stage re-declares only
`USER_UID` and `USER_GID`, so `rzup install rust ${RISC0_TOOLCHAIN}` on line 38 expands to an empty
string either way. Reproduced with the same structure:

```
ARG RISC0_TOOLCHAIN=1.91.1
FROM alpine:3
RUN echo "rzup install rust [${RISC0_TOOLCHAIN}]"   ->  rzup install rust []
```

Worth knowing that #1749 re-declares both `RISC0_VERSION` and `RISC0_TOOLCHAIN` after `FROM`, and
keeps `1.88.0`. The two PRs edit these lines with different values and different structure, so the
merge order decides which version the container installs.

Toby1009 (commented):
Thanks! LGTM. 
I did notice three small things while checking out the branch and running the build. 
They’re all within the scope of this PR, and none of them affect the toolchain decision.

docs/pages/CRISP/introduction.mdx
0xjei, line 1: seems wrong

crates/zk-helpers/src/utils.rs
0xjei, line 1: this might be deleted based on the comment above

crates/zk-helpers/src/circuits/threshold/user_data_encryption/utils.rs
```diff
@@ -22,15 +22,34 @@ use fhe::bfv::{BfvParameters, Ciphertext, PublicKey};
...
+    if ciphertext.len() != 2 {
+        return Err(ZkHelpersUtilsError::UnexpectedCiphertextComponents(
+            ciphertext.len(),
+        ));
+    }
+
     let moduli = params.moduli();
 
-    let ct0is = fhe_poly_to_crt_centered(&ciphertext[0], moduli)?;
-    let ct1is = fhe_poly_to_crt_centered(&ciphertext[1], moduli)?;
```
0xjei, line 33: I don't know if this will maintain the polys coeffs centered when calling again `to_crt()` on them after centering. Worth a double-check.

0xjei (changes requested):

0xjei (approved):
utACK.

We are going to merge this due to time constraints. From now on, we will try to keep things smaller and easier to review. We haven't identified any breaking bugs in there, just a matter of cleaner design

Issue #1794, opened 2026-08-13
docs: 3 broken links remain across circuits, CRISP, and the E3 contract guide


Toby1009:
> **Updated 2026-08-20.** #1795, #1796, #1797 and #1802 fixed 56 of the 59 links. Three remain,
> re-checked against `main` at `3c7172c8`. The original audit, with the full section breakdown, is
> in this issue's edit history.

## 1. `circuits/README.md:85` — a directory that is not there

```
[`wrapper/README.md`](bin/recursive_aggregation/wrapper/README.md)
```

`circuits/bin/recursive_aggregation/` holds `c2ab_fold`, `c3_fold`, `c3_fold_kernel`, `c3ab_fold`,
`c4ab_fold`, `c6_fold`, `c6_fold_kernel`, `decryption_aggregator`, `dkg_aggregator`, `node_fold`,
`nodes_fold`, and `nodes_fold_kernel`. There is no `wrapper/`.

The link is the visible end of a longer drift. The same file names `wrapper/` eleven times,
including the directory tree at `:25`, the C2 description at `:46-50`, and three rows of the circuit
table at `:79-82`. Retargeting the link alone would leave the section describing a layout that does
not exist, so this one needs whoever knows what the wrappers became.

## 2. `examples/CRISP/packages/crisp-contracts/README.md:55` — Greco

```
https://github.com/gnosisguild/interfold/tree/main/circuits/crates/libs/greco
```

404 after redirects. No `greco` directory exists anywhere in the tree. The `gnosisguild` half is not
the cause: that organisation name 301s to `theinterfold`.

## 3. `docs/pages/write-e3-contract.mdx:51` — a pinned commit with a later path

```
https://github.com/theinterfold/interfold/blob/cf50fd6c10f0df8f03eecfd19cedc17b051e72a6/packages/interfold-contracts/contracts/test/MockE3Program.sol#L9
```

404. The commit is reachable and is an ancestor of `main`, but at that revision the file was at
`packages/evm/contracts/test/MockE3Program.sol`. #1592 renamed `packages/evm` to
`packages/interfold-contracts`, and the link took the new path while keeping the old pin. Either
half resolves on its own:

| URL | Status |
| --- | --- |
| `blob/cf50fd6c…/packages/evm/contracts/test/MockE3Program.sol` | 200 |
| `blob/main/packages/interfold-contracts/contracts/test/MockE3Program.sol` | 200 |

Happy to open a PR for 2 and 3 if it is useful.

[...]

PR #1853, opened 2026-08-20
fix(docs): correct the E3 program input hook and a stale source link

Toby1009:
## Problem

`IE3Program` exposes `validate`, `verify` and `publishInput`
(`packages/interfold-contracts/contracts/interfaces/IE3Program.sol:22`, `:37`, `:49`). Five
documents still name a `validateInput` hook, and one of them also attributes it to a caller that
never had it: `Interfold` has no `publishInput`.

Separately, `docs/pages/write-e3-contract.mdx:51` linked `MockE3Program.sol` at commit `cf50fd6c`.
That commit is an ancestor of `main`, but the file was at `packages/evm/` there. #1592 renamed
`packages/evm` to `packages/interfold-contracts`, and the link took the new path while keeping the
old pin, so it 404s.

## Changes

**`examples/CRISP/packages/crisp-contracts/README.md`** described `validateInput`, called by
`Interfold.publishInput`, checking the ciphertext with a Greco library at
`circuits/crates/libs/greco`. None of the three exist. Rewritten against
`CRISPProgram.sol:493-554`: data providers call `publishInput` on the program, which checks the
stage and the input window, resolves eligibility from the census, and verifies a Noir proof over
nine public inputs. The Greco relations that proof checks are implemented in
`circuits/lib/src/core/threshold/user_data_encryption_ct0.nr`, with the witness built in
`crates/zk-helpers`. The list header said two functions above three bullets.

**`examples/CRISP/program/README.md`** tabled `validateInput` and said all three entry points are
called by the Interfold contract. Added a "Called by" column, because that is the part the old text
got wrong.

**`docs/pages/best-practices.mdx:23`** told readers to implement `IE3Program.validateInput`.

**`docs/pages/CRISP/introduction.mdx:110`** named `validateInput()` in the round walkthrough.

**`docs/pages/write-e3-contract.mdx:51`** retargeted to `blob/main`, matching the convention #1802
established. The `#L9` anchor is dropped because line 9 of the current file is blank.

## Verification

| Check | Result |
| --- | --- |
| Both removed links | 404, followed through redirects |
| Both replacement links | 200, no redirect |
| `CRISPProgram.sol:493-554` | line 493 is the `publishInput` signature, line 554 its closing brace |
| `noirPublicInputs` | declared size 9, assigned indices 0–8, contiguous, counted by script |
| `Interfold` has `publishInput` | no match |
| `CRISPProgram` has `validateInput` | no match |
| `pnpm compile` (crisp-contracts) | 46 files, solc 0.8.28 |
| `pnpm build` (docs) | exit 0, `/write-e3-contract` prerendered |
| `format:check`, `check:docs`, `check:invariants` | pass |
| Anchors into either changed page | none, so no inbound link breaks |

On the Greco attribution: `user_data_encryption_ct0.nr:175-178` implements the correct-encryption
relation from [eprint 2024/594](https://eprint.iacr.org/2024/594) directly, together with the
witness range checks, the CRT consistency check on `e0`, and the batched gamma evaluation.
`cargo test -p e3-zk-helpers user_data_encryption` passes 8 tests including
`test_bfv_ciphertext_to_greco`. There is no Noir unit test that exercises the relation itself, so
that part rests on reading the circuit rather than on a test.

## Not included

The sequence diagram in the root `README.md` also names `validateInput`, but four of its five blocks
disagree with the contracts, so it needs redrawing rather than renaming. Sent separately.

Refs #1794

## Summary by CodeRabbit

* **Documentation**
  * Updated CRISP documentation to describe encrypted vote submission through `publishInput`, including submission checks, census-based eligibility, and proof verification.
  * Clarified that verifier selection depends on the round census and distinguished the available verifier contracts.
  * Corrected input-validation guidance to reference the current `IE3Program.publishInput` interface.
  * Refreshed examples with current repository links and clarified contract entry points, callers, and data-provider submission flows.


ctrlc03 (approved):
Thanks

Issue #1811, opened 2026-08-15
fix(support): no gate rebuilds the guest, and the reproducible build works on amd64 only


Toby1009:
> **Updated 2026-08-20 after #1821.** That PR fixed the guest lockfile, brought `RISC0_VERSION` and
> `RISC0_TOOLCHAIN` into stage scope, made `build.rs` derive the guest-builder tag from the
> Dockerfile, and removed the tracked `Elf.sol`. All three original blockers are gone: I ran the
> reproducible build and it now reaches the guest source. The title no longer fits and the first
> section below replaces items 1, 2 and 3a.
>
> **Re-checked 2026-08-21 against `b5811892`.** The reproducible build was run again and behaves
> exactly as described below, down to the same builder digest. Line references into `ci.yml`,
> `releases.yml` and `bump-versions.ts` moved with #1846 and #1848 and are corrected here.

The protocol image ID identifies the guest code that the protocol ciphertext verifier accepts.
Acceptance also requires the protocol-bound journal and the E3 program's separate verification, so
the image ID is not the whole gate. It is the part that identifies code. A reproducible build is
what would connect it back to reviewed source. Without one, a receipt supports "a binary matching
this image ID ran correctly" rather than "the source at commit X ran correctly", and the tracked
`ImageID.sol` carries the value alone, with no revision or build metadata
(`crates/support/contracts/ImageID.sol:17-22`).

#1821 added `docs/pages/verifying-the-compute-provider.mdx` and `pnpm provenance:manifest`, which
record that connection. The page states the same gap in its own words: "Nothing in the repository
reproduces the image ID from the guest source any more, so a match here shows the deployment agrees
with what was committed, not that what was committed is what the source builds."

## The reproducible build now runs, and stops only on an arm64 host

Run against `3c7172c8` from a detached worktree, `RISC0_USE_DOCKER=1 cargo build` in
`crates/support/methods`. The first two blockers are gone:

```
#2 [internal] load metadata for docker.io/risczero/risc0-guest-builder:r0.1.91.1
#3 [build 1/5] FROM docker.io/risczero/risc0-guest-builder:r0.1.91.1@sha256:fafb377a44e1cfca415577c48d2f7012bda99ed36f2fae27f9a663b9fe6048f0
#7 [build 4/5] RUN cargo +risc0 fetch --locked --target riscv32im-risc0-zkvm-elf --manifest-path methods/guest/Cargo.toml
[...]

PR #1854, opened 2026-08-20
fix(support): record the builder image the guest build actually pulls

Toby1009:
## Problem

The provenance manifest records which builder image produced the guest. It named one the build
never pulls.

`risc0-build` resolves the builder tag in three steps
(`risc0-build-3.0.3/src/config.rs:59-70`): `RISC0_DOCKER_CONTAINER_TAG`, then whatever the caller
set on `DockerOptions`, then its own compiled-in `DEFAULT_DOCKER_TAG`.
`crates/support/methods/build.rs:90-95` always sets the second, deriving it from
`ARG RISC0_TOOLCHAIN` in `crates/support/Dockerfile`. The third is therefore unreachable from this
repository — and the third is what the script read, by scanning
`~/.cargo/registry/src/*/risc0-build-3.0.3/src/lib.rs` for the constant.

Same machine, same commit, same source. Only the script differs:

```diff
  "risc0GuestToolchain": "1.91.1",
  "builderImage": {
-   "tag": "risczero/risc0-guest-builder:r0.1.88.0",
-   "digest": null
+   "tag": "risczero/risc0-guest-builder:r0.1.91.1",
+   "digest": "risczero/risc0-guest-builder@sha256:fafb377a44e1cfca415577c48d2f7012bda99ed36f2fae27f9a663b9fe6048f0"
  },
  "unresolved": [
    "guest.elfSha256",
-   "build.builderImage.digest",
    "deployment (pass --rpc and --verifier)"
  ]
```

`digest` was `null` because `docker image inspect` cannot find an image that was never pulled. So
`build.builderImage.digest` could not resolve on any machine, and `complete: true` — what
`docs/pages/verifying-the-compute-provider.mdx` asks a release manifest to reach — was structurally
out of reach. The two remaining entries are conditional rather than broken: `guest.elfSha256` needs
a machine that built the guest, and `deployment` needs `--rpc` and `--verifier`.

## Change

Derive the tag from `ARG RISC0_TOOLCHAIN`, the same source `build.rs` uses, and drop the registry
scan. Two adjacent defects in the same lines went with it, both present before this change:

- **`RISC0_DOCKER_CONTAINER_TAG` is a bare tag, not an image reference.** `docker.rs:145-149`
  formats `risczero/risc0-guest-builder:<tag>` from it. Recording the variable verbatim produced a
  reference docker rejects (`No such image: r0.1.91.1:latest`), leaving the digest unresolved on
  exactly the path an operator uses to pin a builder. The repository prefix now sits outside the
  choice.
- **A whitespace-only ARG produced `risczero/risc0-guest-builder:r0.`** as though it named an
  image. `build.rs:71-77` asserts the trimmed value is non-empty, so the manifest now trims before
  the check and reports the field unresolved instead.

## Verification

Ran the reproducible build, `RISC0_USE_DOCKER=1 cargo build` in `crates/support/methods`:

```
#2 [internal] load metadata for docker.io/risczero/risc0-guest-builder:r0.1.91.1
#3 [build 1/5] FROM docker.io/risczero/risc0-guest-builder:r0.1.91.1@sha256:fafb377a44e1cfca415577c48d2f7012bda99ed36f2fae27f9a663b9fe6048f0
```

That is the tag and the digest the manifest now records.

Both tag paths were run and agree:

| Run | Recorded tag |
| --- | --- |
| No override | `risczero/risc0-guest-builder:r0.1.91.1` |
| `RISC0_DOCKER_CONTAINER_TAG=r0.1.91.1` | `risczero/risc0-guest-builder:r0.1.91.1` |

The four toolchain inputs behave as intended: `"1.91.1"` gives the tag, and `""`, `"   "` and
`null` each give a null tag rather than a partial one.

`tsc --noEmit`, `eslint` and `prettier --check` pass on the file.

Refs #1811

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved builder image selection to align with the configured guest toolchain.
  * Added support for whitespace-only toolchain values and clearer fallback behavior when no toolchain is available.
  * Updated incomplete manifest warning messaging for improved clarity.


ctrlc03 (approved):
Thanks

PR #1867, opened 2026-08-22
fix: event polling to fetch and emit new events

leonschh:
### Problem

`packages/interfold-sdk/src/events/event-listener.ts` had a bug in `pollForEvents():` it only advanced `lastBlockNumber` without fetching or emitting logs. As a result, `startEventPolling()` could never trigger event callbacks.

This contradicts the documented behavior in [`docs/pages/sdk.mdx`](https://github.com/theinterfold/interfold/blob/main/docs/pages/sdk.mdx), which states that `sdk.startEventPolling()` can be used to poll logs over HTTP when WebSockets are unavailable.

### Fix
- Fetch historical events for each subscribed event type between the last and current block.
- Emit fetched logs through the existing `emit()` pipeline.
- Extract `buildEvent()` to share event construction between polling and WebSocket watchers.
- Advance `lastBlockNumber` only after successful event processing.

The existing event tests in [`packages/interfold-sdk/tests/events.test.ts`](https://github.com/theinterfold/interfold/blob/main/packages/interfold-sdk/tests/events.test.ts#L87-L91) (Lines 87-91) confirm that emitted events are delivered to registered callbacks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved contract-event monitoring to consistently normalize and emit events.
  * Polling now captures historical events across newly detected blocks, reducing the risk of missed event notifications.


ctrlc03 (changes requested):
the code does not compile

PR #1876, opened 2026-08-24
fix(docs): repair ciphernode getting started links

jfschwarz:
Replace the four relative links in the Ciphernode Operators Getting Started section (they rendered as dead links)


ctrlc03 (approved):
Thanks

PR #1910, opened 2026-09-01
chore(ci): move trusted workflows to RunsOn

auryn-macmillan:
Quick PR to move the github workflows to runs-on to see if they go faster there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,12 +36,16 @@ env:
 permissions:
   contents: read
 
-# Fork pull requests run untrusted code, so they stay on GitHub-hosted runners.
+# The migrated jobs use RunsOn only for protected branch events.
+# Pull requests keep their previous runner targets, including enclave-ci where configured.
 # Each RunsOn label embeds run id, attempt, and job ID so jobs never share runners.
 jobs:
   detect_changes:
     timeout-minutes: 5
-    runs-on: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository && 'ubuntu-latest' || format('runs-on={0}-{1}-{2}/runner=2cpu-linux-x64', github.run_id, github.run_attempt, 'detect_changes') }}
+    runs-on:
+      ${{ github.repository == 'theinterfold/interfold' && github.event_name != 'pull_request' && (github.ref == 'refs/heads/main' ||
+      github.ref == 'refs/heads/dev') && format('runs-on={0}-{1}-{2}/runner=4cpu-linux-x64/ram=16', github.run_id, github.run_attempt,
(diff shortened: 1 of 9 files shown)


.github/workflows/releases.yml
```diff
@@ -201,7 +202,7 @@ jobs:
 
   publish-npm-packages:
     name: Publish NPM Packages
-    runs-on: ubuntu-latest
+    runs-on: runs-on=${{ github.run_id }}/runner=2cpu-linux-x64
     needs: [validate-and-prepare, release-candidate-gate]
     permissions:
       contents: read
```
hmzakhalid, line 207: this would fail since npm doesnt allow self hosted runner, only OIDC publisher, it supports github hosted runners.

.github/workflows/ci.yml
```diff
@@ -39,7 +39,7 @@ permissions:
 jobs:
   detect_changes:
     timeout-minutes: 5
-    runs-on: 'ubuntu-latest'
+    runs-on: runs-on=${{ github.run_id }}/runner=2cpu-linux-x64
```
hmzakhalid: since this is a public repo, fork PRs can now run arbitrary code on our self hosted AWS runner. these labels also default to the production RunsOn environment. can we keep fork PRs on GH hosted runners or isolate them properly first?

.github/workflows/ci.yml
```diff
@@ -172,7 +172,7 @@ jobs:
     needs: [detect_changes]
     if: needs.detect_changes.outputs.rust_unit_tests == 'true'
     timeout-minutes: 20
-    runs-on: 'ubuntu-latest'
+    runs-on: runs-on=${{ github.run_id }}/runner=2cpu-linux-x64
```
hmzakhalid: looks like all these jobs use the same `github.run_id` label. RunsOn says concurrent jobs can steal each other’s runners when the labels match. can we add the job name and `github.run_attempt` plus `strategy.job-index` for matrix jobs?

.github/workflows/releases.yml
auryn-macmillan: publish-npm-packages pinned to ubuntu-latest.

.github/workflows/ci.yml
auryn-macmillan: every PR-triggered job (7 files) now evaluates `github.event_name == 'pull_request' && head.repo.full_name != repository` and routes fork PRs to `ubuntu-latest` before any step starts. Non-PR events (push/cron/dispatch) are unaffected by the guard. Verified live: this PR's own check ran on `labels: ["ubuntu-latest"], runner_group: "GitHub Actions"`.

.github/workflows/ci.yml
auryn-macmillan: every RunsOn label is now a unique string.

PR #1928, opened 2026-09-08
fix: medium audit issues [skip-line-limit]

ctrlc03:
## Summary by CodeRabbit

* **New Features**
  * Added operator-based claiming and tracking for held rewards and slash distributions.
  * Added reward-status and held-balance visibility.
  * Added failure-reclassification reporting when committee expulsions change the failure cause.

* **Bug Fixes**
  * Finalized committee collateral remains locked until the accusation window closes.
  * Corrected penalty redistribution so each penalty excludes only its original target.
  * Improved reward accounting and claim event reporting.
  * Added safeguards against data-availability provider changes during active rounds.

* **Documentation**
  * Documented updated lifecycle, refund, slashing, and data-availability behavior.


packages/interfold-contracts/contracts/lib/InterfoldLifecycle.sol
```diff
@@ -495,6 +496,104 @@ library InterfoldLifecycle {
...
+        // rather than a failure the caller has to catch.
+        //
+        // Settlement snapshots the payer from the reason. After that the
+        // distribution is fixed, so a late correction must not disagree with
+        // the amounts already credited.
+        if (
+            IE3RefundManager(refundManager)
+                .getRefundDistribution(e3Id)
+                .calculated
+        ) return;
```
hmzakhalid: this still leaves #4 open if the member adds `processE3Failure()` to the front-run. even with valid, already-executable expelling proposals that would take the committee below H, they can do `markE3Failed → processE3Failure → executeSlash`. this guard then keeps the requester-paid result.

the new “settlement makes the recorded reason final” test covers that ordering. with its allocation, the requester still gets 45% of service escrow instead of the 100% supplier-failure refund. penalties still execute, but that doesn't correct who paid.

i get why settled payouts must stay final. can we define a bounded pre-settlement rule for those pending/executable proposals before fixing the payer? otherwise let's call this a partial mitigation and get explicit risk acceptance, rather than say attribution is order independent. this affects the [fault-attribution and settlement invariants](https://github.com/theinterfold/interfold/blob/c0aed74f169659a754aacae13afad1954d6e33c9/agent/INVARIANTS.md#slashing-and-failure-settlement).

packages/interfold-contracts/contracts/lib/RegistrySortitionLib.sol
```diff
@@ -373,6 +399,47 @@ library RegistrySortitionLib {
...
+            revert ICiphernodeRegistry.CommitteeObligationsAlreadyReleased(
+                e3Id
+            );
+        if (
+            e3Stage != IInterfold.E3Stage.Complete &&
+            e3Stage != IInterfold.E3Stage.Failed
+        ) revert ICiphernodeRegistry.E3NotTerminal(e3Id);
+        if (stage != ICiphernodeRegistry.CommitteeStage.Finalized) return;
+        uint64 submissionDeadline = slashingManager
+            .accusationSubmissionDeadline(e3Id);
```
hmzakhalid: can we include the rollout steps for this new dependency? `SlashingManager` is deployed directly in `scripts/protocol/deployContracts.ts`, and `safeProxyUpgrade.ts` only handles the four proxy targets.

if we upgrade the registry but leave the existing manager in place, finalized `releaseCommittee()` calls will revert here because that manager doesn't have `accusationSubmissionDeadline()`. regenerating the ABI won't update it.

we need the replacement manager wired through the paused/drained dependency graph, plus an upgrade-compatibility check before requests resume. this is the [one coherent dependency generation invariant](https://github.com/theinterfold/interfold/blob/c0aed74f169659a754aacae13afad1954d6e33c9/agent/INVARIANTS.md#e3-request-and-committee-selection). the new-deployment fixtures don't establish that the existing deployment can make this transition.

hmzakhalid (dismissed):

packages/interfold-contracts/contracts/E3RefundManager.sol
```diff
@@ -890,20 +886,83 @@ contract E3RefundManager is IE3RefundManager, Ownable2StepUpgradeable {
...
+        ).getActiveCommitteeNodes(e3Id);
+        mapping(address target => uint256 amount)
+            storage buckets = _heldSlashFrom[e3Id][operator];
+        for (uint256 t = 0; t < nodes.length; t++) {
+            uint256 amount = buckets[nodes[t]];
+            if (amount == 0) continue;
+            buckets[nodes[t]] = 0;
+            _reshareHeldSlashBucket(e3Id, nodes, operator, nodes[t], amount);
+        }
+    }
```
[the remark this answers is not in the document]
hmzakhalid: yeah, traced this too. concrete case on a failed, settled round with A/B/C:

1. submit valid expelling proposals against A and B while all three are still active, then let them become executable.
2. expel A and settle its 50-unit ticket penalty. B and C each hold 25.
3. expel B. `_resolveExpulsionOutcome` clears B's `heldSlash`, but this loop only sees active C, so it never reads `_heldSlashFrom[e3Id][B][A]`.

B's 25 from A is now stranded. after B's own 50-unit penalty settles, C has 75 instead of 100. the missing 25 stays in `_tokenLiability` with no payable allocation.

can we enumerate the stored bucket keys independently of active membership, for both redistribution and claim cleanup, and add this two-expulsion case? also assert that the liability equals the remaining allocations, not just `balance >= liability` — a solvent contract can still strand these funds.

this breaks the [order-independent reward reallocation invariant](https://github.com/theinterfold/interfold/blob/c0aed74f169659a754aacae13afad1954d6e33c9/agent/INVARIANTS.md#slashing-and-failure-settlement). this sequence is from the source trace; i haven't run it locally.

PR #1939, opened 2026-09-14
feat: crash-tolerant dkg [skip-line-limit]

hmzakhalid:
## Scope

- Recover DKG documents and proof work after a restart.
- Let the elected DKG aggregator collect readiness reports and announce one canonical H-member roster when early DKG members are absent. Reuse the existing aggregator failover path instead of adding a second leader system.
- Verify the selected roster in the recursive proof. Index selected proof rows by original party ID, retain the required N-wide C3 witness slots when a C0 recipient is absent, and build valid C4 witnesses for nodes outside the selected roster.
- Persist commitment checks, C4 collection, and node-fold state so a restart resumes from durable inputs.
- Change Small to N=19, H=14, T=9 and rebuild all affected circuits and verifiers.
- Give CRISP proposals fixed voting start and end dates, followed by a separate Avail finalization window.
- Fix the Faucet token binding, deterministic bonded-voter census, late replay handling, large work-plan persistence, and process, RPC, and peer-liveness defects found during the Sepolia rehearsals.

## Verification

- All checks pass at `e96c8cd4c`, including Rust unit and integration tests, contract tests, the full six-pair circuit release matrix, zk-prover tests, SDK tests, and CRISP end-to-end tests.
- A previous Sepolia Small N=19/H=10 round completed end to end: DKG published the key, Avail finalized the ballot, Boundless fulfilled the computation, the committee decrypted the result, and CRISP reported `Finished` with tally `1-0`.
- A fresh Sepolia Small N=19/H=14 round is running on 21 dedicated hosts with the exact candidate binary and circuit archive. The selected 19 formed a complete transport and GossipSub mesh, received all 19 C0 keys, and entered proof generation with zero process restarts.
- A local insecure Small omission run disconnected party 0 after C0. The remaining nodes selected a non-prefix roster, published one key, and recovered the expected plaintext with real C0-C7 leaf proofs.
- A clean local Minimum run restarted a selected member during C4 with the production aggregator timeout. It published the key and recovered the expected tallies `15`, `14`, and `9`.
- All Minimum, Micro, and Small circuit pairs for insecure-512 and secure-8192 were rebuilt. Both generated verifier types match each pair.

## Hygiene

- The committed diff contains no workstation paths, server addresses, credentials, branch or PR references, debug markers, rehearsal logs, or one-off deployment helpers.
- Fault controls remain in the integration harness. Production binaries reject proof skipping. Synthetic DKG timing is available only with the test-helper feature.
- Fixed values are named protocol or safety bounds. Generated verifier constants and deployment addresses remain in their canonical generated or configuration files.
- CI paths are repository-local or runner-local scratch paths and do not depend on a developer machine.
- The title keeps the repository `[skip-line-limit]` marker because this is the requested consolidated change.

## Launch constraints

Small is now N=19, H=14, T=9. DKG can omit up to five members before the roster is accepted, and threshold decryption can lose up to four of the 14 key holders after key generation. The launch flow does not replace an accepted roster member who disappears permanently before sending C4; it relies on the implemented crash-and-restart recovery for that stage.

This is the trusted-operator launch design, not a Byzantine-safe asynchronous DKG agreement protocol. Protocol v4 is intentionally not backward compatible with persisted v3 state; test nodes must clear protocol-scoped state before starting this binary.

## Coordinated companion PRs

- Voting plugin: theinterfold/crisp-aragon-plugin#46
- DAO app: theinterfold/crisp-aragon-plugin-app#87
- Fleet automation: theinterfold/interfold-inf#1

Deploy a new CRISPProgram first. Then deploy the plugin and DAO, followed by the app. Promote all six circuit pairs to `circuit-artifacts` immediately before or with the merge.


hmzakhalid:
Validation update: the first full-proof insecure Small (N=19/H=10) offline-after-C0 run reached a shared non-prefix roster and completed 8 NodeDkgFold proofs, but the test harness stopped waiting for PublicKeyAggregated at 5,000s while its synthetic DKG window was 7,200s. Commit e1567873 adds test-only observation-timeout overrides. A new run is underway with a 10,800s DKG window and 14,400s observation budget. This is not yet a passing Small result; the PR stays draft. The previous head completed all CI checks successfully.

PR #1977, opened 2026-09-20
fix(runtime): harden canonical event ingestion [skip-line-limit]

hmzakhalid:
## Summary

- preserve separate EVM log occurrences without changing persisted logical event IDs
- admit every durable event to snapshot batching before deduplication and domain fanout
- reject stale snapshot writes that would replace newer aggregate state
- default EVM ingestion to one confirmation, including through local RPC proxies
- tolerate short provider lag when a newly announced block is not immediately readable
- detect existing sortition projection corruption during node validation

### Snapshot projection corruption

EventStore accepted separate chain-log occurrences with the same decoded payload. EventBus then deduplicated them using only the payload-derived event ID. Because SnapshotBuffer was downstream of that deduplication, a suppressed occurrence could leave a sequence gap. The router waited for an exact predecessor, so a delayed older batch could later replace newer actor state while the newer aggregate cursor remained persisted.

### Missed EVM requests behind an RPC pool

A localhost URL automatically selected zero-confirmation ingestion. When that URL fronted several providers, one provider could announce a subscription head while another still returned no block for that height. Timestamp lookup then rejected the live log and reconnected, creating a gap in which a short-lived E3 request could be missed.

### Event delivery and persistence

- add a local delivery identity that includes the EVM chain, block, and deterministic log timestamp/index
- centralize mandatory snapshot admission at the EventBus boundary before domain deduplication
- cover live, replay, source-bus, and forked-bus delivery paths with the same admission rule
- fail startup if snapshot admission cannot be registered, and signal runtime persistence failure if admission stops
- schedule all older open snapshot batches when a newer sequence arrives
- atomically reject stale contextual writes in both Sled and in-memory stores

### EVM ingestion

- default each chain to one confirmation unless configuration explicitly selects zero
- preserve the configured confirmation depth when startup resumes from snapshot cursors
- use a log-provided block timestamp when available
- retry a temporarily missing block with bounded backoff before rejecting a live log
- keep incomplete initial historical sync fatal; only live ingestion reconnects for canonical backfill

### Validation and recovery

- compare persisted sortition state with source events through each aggregate's persisted snapshot cursor
- report missing and unexpected operator addresses when validation finds a mismatch
- document the controlled rescan path for data that was already corrupted before this fix

## Compatibility

This PR does not change persisted event schemas, storage schemas, circuits, or contracts. Existing healthy node data remains readable. The runtime prevents new stale overwrites, but it does not invent missing state for an already-corrupted node.

The changes preserve the commit-before-dispatch and replay-equivalence requirements in `agent/invariants/03_ACTOR_RUNTIME.md`.

## Verification

- 490 scoped library tests passed across `e3-events`, `e3-data`, `e3-config`, `e3-evm`, `e3-sync`, `e3-entrypoint`, and `e3-ciphernode-builder`
- `cargo check --workspace --all-targets`
- `cargo fmt --all -- --check`
- `pnpm lint`
- `pnpm check:license`
- `pnpm check:committee`
- `pnpm check:docs`
- `pnpm check:addresses`
- `pnpm check:invariants`
- `pnpm check:verifiers`


hmzakhalid:
Review follow-up completed in `b3a955d8a`.

- Clarified that incomplete initial historical sync is fatal, while live ingestion reconnects for canonical backfill.
- Moved mandatory snapshot admission to the EventBus boundary so normal, replay, source-bus, and forked-bus events all use the same ordering.
- Added source/fork/replay coverage.
- Preserved configured confirmation depth when resuming from snapshot cursors.
- Scoped sortition validation to events covered by each persisted snapshot cursor.
- Included chain identity in EVM delivery deduplication.

The full workspace check and 490 scoped library tests pass locally.

PR #1979, opened 2026-09-21
feat: introduce insecure preset & benches for l-bfv [skip-line-limit]

0xjei:


0xjei:
got a review from @ctrlc03 & fixed! all good

